PR #76, opened 2014-12-09
Adds a new parameter show namespaces

nischayn22:
This parameter can be used to disable the namespace in the results output
by the format.


Fannon:
I'm not a 100% sure, but this could be solved in a more flexible and generic way with `{{DISPLAYTITLE:{{PAGENAME}}}}` (see https://www.mediawiki.org/wiki/Extension:Semantic_Title) and proper support for this in all result formats. As of SMW 2.4, it should already be working with most formats.

JeroenDeDauw:
Let's close this then

PR #91, opened 2015-02-26
Fix class ref in slideshow API

JeroenDeDauw:


s7eph4n:
Thanks, @JeroenDeDauw !

Looks good to me, but I do not have a test setup to verify. Suggest to merge and ask @jaideraf and/or @mitchelln to confirm the issue can be closed.

JeroenDeDauw:
It's obvious enough the ref was broken and now is not

jaideraf:
@s7eph4n I confirm: slideshow format works now.
@JeroenDeDauw: thanks for the fix.

mitchelln:
I will also test, but is sounds like it has been resolved. Good work!

mitchelln:
Confirmed fixed.

JeroenDeDauw:
I tagged the rel earlier today, so you can do a composer update to get it

PR #102, opened 2015-04-25
minor fix of tooltip for month view

bogota:
- If the length of the event description <= self.defaults.descriptionLength, we shouldn't process it to be shortened.

- The tooltip failed to display event descriptions having no space characters within. It matters when we deal with CJK texts.


JeroenDeDauw:
Thanks for the contribution! Can you put this new logic into a new `getTruncatedSentence` function, so the one you are modifying does not get bigger?

mwjames:
@bogota I haven't tested it but it looks sane therefore :+1: .

PR #115, opened 2015-07-24
[Calendar] added startmonth and startyear parameter

gesinn-it:
I needed SRF Calendar to show a specific month/year instead of the current one. So I implemented two new parameter: startmonth and startyear. Both defaults to current values if not specified.

### DIFF
--- a/formats/calendar/SRF_Calendar.php
+++ b/formats/calendar/SRF_Calendar.php
@@ -14,6 +14,8 @@ class SRFCalendar extends SMWResultPrinter {
 	protected $mTemplate;
 	protected $mUserParam;
 	protected $mRealUserLang = null;
+	protected $mStartMonth;
+	protected $mStartYear;
 
 	protected function setColors( $colorsText ) {
 		$colors = array();
@@ -32,6 +34,8 @@ class SRFCalendar extends SMWResultPrinter {
 
 		$this->mTemplate = trim( $params['template'] );
 		$this->mUserParam = trim( $params['userparam'] );
+		$this->mStartMonth = trim( $params['startmonth'] );	// startmonth is initialized with current month by default
+		$this->mStartYear = trim( $params['startyear'] );	// startyear is initialized with current year by default
 
 		if ( $params['lang'] !== false ) {
 			global $wgLang;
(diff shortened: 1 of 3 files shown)


gesinn-it:
Can someone please have a look at this and verify if it can be merged. Thx, Alexander.

formats/calendar/SRF_Calendar.php
```diff
@@ -357,7 +367,12 @@ function displayCalendar( $events ) {
 		}
 
 		$cur_month = self::intToMonth( $cur_month_num );
-		$cur_year = date( 'Y', time() );
+
+		if ( is_numeric( $this->mStartYear ) && ( intval( $this->mStartYear ) == $this->mStartYear ) ) {
```
mwjames: Since you removed the validation on behalf of  `time()` what happens when `$this->mStartYear` is  `-9000`? Does it still work without crashing due to out of bound limits?

gesinn-it: Hi mwjames, thanks for having a look into this. I did a test with mStartYear = -2000 and it works fine (not crashing the calendar). I assume that this is interpreted as 2000 years before the Nativity. If no year is specified it still sets $cur_year = date( 'Y', time() );

formats/calendar/SRF_Calendar.php
```diff
@@ -357,7 +367,12 @@ function displayCalendar( $events ) {
...
+			$cur_year = $this->mStartYear;
+		} else {
+			$cur_year = date( 'Y' );
```
mwjames: Is `$cur_year = date( 'Y' );` === `$cur_year = date( 'Y', time() );`?

formats/calendar/SRF_Calendar.php
```diff
@@ -348,7 +352,13 @@ function displayCalendar( $events ) {
...
 		// be in a different year), the number of days in the current,
 		// previous and next months, etc.
-		$cur_month_num = date( 'n', time() );
+
+
+		if ( is_numeric( $this->mStartMonth ) && ( intval( $this->mStartMonth ) == $this->mStartMonth ) && $this->mStartMonth >= 1 && $this->mStartMonth <= 12 ) {
+			$cur_month_num = $this->mStartMonth;
+		} else {
+			$cur_month_num = date( 'n' );
+		}
```
mwjames: `$cur_month_num = date( 'n' );` === `$cur_month_num = date( 'n', time() );`?

formats/calendar/SRF_Calendar.php
gesinn-it: I'd say so.

Returns a string formatted according to the given format string using the given integer timestamp or the current time if no timestamp is given. In other words, timestamp is optional and defaults to the value of time(). [1]

[1] http://php.net/manual/en/function.date.php

mwjames: :+1: 

I haven't checked it, I just wanted to make sure it doens't return with something unexpected.

mwjames:
Thanks.

@kghbln Does it need to be mentioned in the README?

gesinn-it:
@mwjames Thanks.

Please, let me know if I shall update [1] or if this will be done after the next release.

[1] https://www.semantic-mediawiki.org/wiki/Help:Calendar_format

kghbln:
> Does it need to be mentioned in the README? 

@mwjames Yes, it should since it is an enhancement. Also a little more to show. So I did with commit https://github.com/SemanticMediaWiki/SemanticResultFormats/commit/56c5f6151648798602392f7235c99cc415935021 Also amended the info while I was at it. I think after the next sync with the WMF-repo by @JeroenDeDauw we could do a release.

@gesinn-it Will do. I believe the docinfo table will update automatically once the version is updated on swm.o. So just a little note is needed.

JeroenDeDauw:
Just did a sync

kghbln:
@gesinn-it Thanks for the enhancement and @mwjames Thanks for reviewing it and @JeroenDeDauw Thanks for syncing. Everything in fluff now. :)

PR #118, opened 2015-08-25
Update SRF to use Contexts

hexmode:
I ran into some problems where various bits here were undefined when
running SMW's rebuildData.

### DIFF
--- a/formats/calendar/SRF_Calendar.php
+++ b/formats/calendar/SRF_Calendar.php
@@ -262,12 +262,14 @@ class SRFCalendar extends SMWResultPrinter {
 	}
 
 	function displayCalendar( $events ) {
-		global $wgOut, $wgParser, $wgRequest;
+		global $wgParser;
 		global $srfgFirstDayOfWeek;
 
+		$context = RequestContext::getMain();
+		$request = $context->getRequest();
 		$wgParser->disableCache();
 
-		$wgOut->addLink( array(
+		$context->getOutput()->addLink( array(
 			'rel' => 'stylesheet',
 			'type' => 'text/css',
 			'media' => 'screen, print',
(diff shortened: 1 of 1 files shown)


JeroenDeDauw:
Thanks for the contribution! Had a quick look and did not see anything that should be changed.

A bit dubious about some of the style changes though; mainly that if code is nested too much, it's better to split the functions to avoid that then to split the lines themselves. Anyway, that is not a big deal compared to the style issues already present, and you certainly fixed some as well.

Is this supposed to still work with MW 1.19?

hexmode:
> Is this supposed to still work with MW 1.19?

The interface (RequestContext::getMain()) [was included in 1.19](https://git.wikimedia.org/commit/mediawiki%2Fcore/8dfa97ff0a6fd4bb506394a65b5653a0547243b3).

hexmode:
Jeroen De Dauw writes:

> A bit dubious about some of the style changes though; mainly that if code is nested too much, it's better to split the functions to avoid that then to split the lines themselves.

Yes, I got some warnings about length and the like.  Easy problems
first.

-- 
Mark A. Hershberger
NicheWork LLC
717-271-1084

JeroenDeDauw:
@hexmode did you manually test that things still work after this change? We do not have many tests in this extension, so breakage might sail right through the CI.

hexmode:
I have tested it lightly using an #ask query with format=calendar.

JeroenDeDauw:
Ok, thanks!

hexmode:
Note that I originally discovered this problem when these errors popped up:
    2015-09-16 15:28:03 refreshLinks Course_calendar (id=2548866) STARTING
    PHP Fatal error:  Call to a member function getLocalURL() on a non-object in .../extensions
    /SemanticResultFormats/formats/calendar/SRF_Calendar.php on line 404

JeroenDeDauw:
This needs to be fixed before the next release, preferably ASAP, else we'll need to revert.

hexmode:
If I can reproduce it today, I'll fix it.

PR #133, opened 2015-10-28
Alpha version of SRF_Word.php to discuss -cave has one change in Math not related ...

### LINKED ISSUE #114: add word format
As an enhancement I'd like to add a word format. I have started by forking SemanticResultFormats and made some progress see

 https://github.com/BITPlan/SemanticResultFormats/blob/6ac8769ea7ba5b22abfaa504afc015cbe3dc474c/formats/word/SRF_Word.php

I am having trouble looping over the DataItems and Values - my example has some Date values in it and they are not catched in the field loop yet.

WolfgangFahl:
As dicussed this Pull-Request is for review only not to merge yet.

Fixes #114


mwjames:
This has been open since Oct 28, 2015, so what is the plan here?

WolfgangFahl:
I would love to work on this. During the SMWCon in Barcelona it was pointed out to me that code would only be accepted with Unit Tests. I have no clue how to write those unit tests properly and since  https://github.com/SemanticMediaWiki/SemanticMediaWiki/blob/master/tests/phpunit/includes/queryprinters/ResultPrintersTest.php doesn't help me in this case I got stuck.

I still think https://www.semantic-mediawiki.org/wiki/1st_SMW_webinar_for_developers is a good idea to make sure issues like this do not get stuck.

mwjames:
> https://github.com/SemanticMediaWiki/SemanticMediaWiki/blob/master/tests/phpunit/includes/queryprinters/ResultPrintersTest.php doesn't help me in this case I got stuck.

What you can do is to isolate the function that can be tested independently, for example all the file operations and ob_start ... operations should be handle by a separate class with a "slim" interface. 

Here are some recent issue examples:

- 297
- 314
- 338

mwjames:
> This has been open since Oct 28, 2015, so what is the plan here?
>  mwjames commented on Nov 4, 2017

I ask again, do we expect any progress here, or should we close this. It should be noted that https://github.com/SemanticMediaWiki/SemanticResultFormats/pull/445 was merged and might allow for some inspiration.

> . During the SMWCon in Barcelona it was pointed out to me that code would only be accepted with Unit Tests.

Yes, this policy has been in place for some time now.

> I have no clue how to write those unit tests properly and since 

There are a couple of examples in the repository that can be used as blueprint, if there are concrete questions I may answer them depending on the available time.

General comment on the `SRF_Word.php`:
- the class is monolith  and should be split according the responsibilities such as handling file contents, formatting values etc.
- Isolate responsibilities by class which allows them to be tested separately because the way `SRF_Word.php` is structured, you cannot write any meaningful unit test

WolfgangFahl:
The project i was using this result format for is not active any more. So currently I won't invest extra time in this. For the time being I am opting for "WON'T FIX" until someone is willing to spend more time with this.

mwjames:
> The project i was using this result format for is not active any more. So currently I won't invest extra time in this. For the time being I am opting for "WON'T FIX" until someone is willing to spend more time with this.

Thanks for the feedback.

PR #139, opened 2016-01-08
Fix for running composer locally

cicalese:
If another composer-installed extension (e.g. Semantic Extra Special Properties) installs Semantic MediaWiki in its local extension vendor directory, Semantic MediaWiki will be autoloaded from there, but the files for this extension will never be autoloaded (because SMW_VERSION will already be defined). Perhaps there is a better solution, but with this change, if composer is run in the SemanticResultFormats directory to create its local vendor directory, the code will now be sure to autoload the files for this extension.

### DIFF
--- a/SemanticResultFormats.php
+++ b/SemanticResultFormats.php
@@ -23,7 +23,7 @@ if ( version_compare( $GLOBALS['wgVersion'], '1.19c', '<' ) ) {
 	throw new Exception( 'This version of Semantic Result Formats requires MediaWiki 1.19 or above; use SRF 1.7.x or SRF 1.6.x for older versions.' );
 }
 
-if ( !defined( 'SMW_VERSION' ) && is_readable( __DIR__ . '/vendor/autoload.php' ) ) {
+if ( is_readable( __DIR__ . '/vendor/autoload.php' ) ) {
 	include_once( __DIR__ . '/vendor/autoload.php' );
 }
 


mwjames:
> If another composer-installed extension (e.g. Semantic Extra Special Properties) installs Semantic MediaWiki in its local extension vendor directory, 

Why would that happen?

As far as I can recall, the installation instructions for both SMW/SRF are explicitly pointing to use the MW root by either adjusting `composer.json` or the MW specific local.json thing. Why would SRF (or for that matter SESP) install SMW in its own directory and thereof have to handle a vendor file locally?

cicalese:
I was asked to create a MediaWiki-vagrant role for SMW (see https://phabricator.wikimedia.org/T99858). MediaWiki-vagrant is used primarily for experimentation and development. I thought it would also be useful to create MediaWiki-vagrant roles for some helpful companion extensions to SMW. This work will allow a broad community of folks to easily gain exposure to the SMW suite of extensions in an experimental environment. The composer installation approach for extensions in MediaWiki-vagrant is to run composer in the extension directories, not at the root directory. With this simple fix, running composer in the extension directories works great - again, primarily for experimentation and development. Modifying MediaWiki-vagrant to run composer in the root directory for extensions would be a considerably more complicated endeavor for a variety of reasons.

Cindy



Sent with Good (www.good.com)
________________________________

mwjames:
> I was asked to create a MediaWiki-vagrant role for SMW (see https://phabricator.wikimedia.org/T99858). 

I'm not having an issue with this change, just need some clarification as I pointed out in [0].

[0] https://github.com/SemanticMediaWiki/SemanticExtraSpecialProperties/pull/57

cicalese:
Yes, I was just including the background here as well to give context for the request. Thanks for providing the link to the other patch. I was not on a device where I could easily do so before.

cicalese:
FYI, I believe I will have a similar issue if I attempt to create MediaWiki-vagrant roles for Maps and SemanticMaps.

cicalese:
Thank you!

PR #144, opened 2016-01-18
Added displaytitle parameter, for rudimentary DISPLAYTITLE / Semantic Title Support

Fannon:
This patch will add support for SemanticTitle for the filtered format. It is far from optimal, but as the displaytitle is currently not available in the QueryResult object this is what I came up with.

When activated with displaytitle=true (by default the patch will do nothing), a map of title -> displaytitle will be built and used to assign the Filter Checkboxes the correct label.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -1,5 +1,5 @@
 {
-	"name": "fannon/semantic-result-formats",
+	"name": "mediawiki/semantic-result-formats",
 	"type": "mediawiki-extension",
 	"description": "Additional result formats for Semantic MediaWiki queries",
 	"keywords": [
--- a/composer.json
+++ b/composer.json
@@ -1,5 +1,5 @@
 {
-	"name": "mediawiki/semantic-result-formats",
+	"name": "fannon/semantic-result-formats",
 	"type": "mediawiki-extension",
 	"description": "Additional result formats for Semantic MediaWiki queries",
 	"keywords": [
--- a/formats/Filtered/SRF_Filtered.php
+++ b/formats/Filtered/SRF_Filtered.php
@@ -248,6 +248,7 @@ class SRFFiltered extends SMWResultPrinter {
 				', "filterdata" : ' . json_encode( $filterData ) .
 				', "sorthandlers" : ' . json_encode( array() ) .
 				', "sorterdata" : ' . json_encode( array() ) .
+				', "displaytitle" : ' . json_encode( $this->mParams['displaytitle'] ) .
 //				', "sorterhandlers" : ' . json_encode( $sorterHandlers ) .
 //				', "sorterdata" : ' . json_encode( $sorterData ) .
 				'}};'
@@ -292,6 +293,14 @@ class SRFFiltered extends SMWResultPrinter {
 			'name' => 'filter position',
 			'message' => 'srf-paramdesc-filtered-filter-position',
 			'default' => 'top',
+			// 'islist' => false,
+		);
+
+		$params[] = array(
+			'type' => 'boolean',
+			'name' => 'displaytitle',
+			'message' => 'srf-paramdesc-filtered-displaytitle',
+			'default' => false,
 			// 'islist' => false,
 		);
 
--- a/formats/Filtered/libs/ext.srf.filtered.js
+++ b/formats/Filtered/libs/ext.srf.filtered.js
@@ -29,6 +29,14 @@
 					var title = $(this).attr('title');
 					if (title) {
 						data['data']['displaytitle-map'][title] = $(this).text();
+
+						// Handle namespaces containing spaces/underscores
+						if (title.indexOf(':') > -1) {
+							titleArray = title.split(':');
+							titleArray[0] = titleArray[0].split(' ').join('_');
+							title = titleArray.join(':');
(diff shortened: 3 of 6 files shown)


composer.json
```diff
@@ -1,5 +1,5 @@
 {
-	"name": "mediawiki/semantic-result-formats",
+	"name": "fannon/semantic-result-formats",
```
JeroenDeDauw: Think you accidentally left this in

Fannon: No, I just added it to test it with composer, sorry !

gesinn-it:
It would be cool if we could merge this PR. It is not ideal but provide a good way to use Filtered together with Semantic Title. Without this PR, Filtered could not be used together with Semantic Title. I know, that @cicalese is trying to get a more sophisticated general solution merged. But this seems to take some more cycles. In the meanwhile @Fannon 's patch could do a really good job in my opinion.

gesinn-it:
@mwjames , @JeroenDeDauw can you please have a look, if this PR is ready to be merged. Many thanks.

mwjames:
I'd be happy to merge once @s7eph4n says that the changes do contain any issues.

gesinn-it:
@s7eph4n can you please have a look if you have any objections to merge this PR

s7eph4n:
@mwjames Did not test it, but it looks good. There is a call to console.log() still in, but else it should be fine.

s7eph4n:
Sorry, console.info() I meant. And it really should go, it is non-standard. See https://developer.mozilla.org/en-US/docs/Web/API/Console/info.

Fannon:
Yes, you're right! Anything using the console.object should not be in production code!

Fannon:
This remomoves the console.info and also fixes some issues with namespaces containing spaces

Fannon:
I've just fixed one more issue with HTML Entity encodings, causing this patch to miss the Semantic Title replacement in some reare cases.

Fannon:
@s7eph4n Do you think, this would be ready for a merge?

cicalese:
@Fannon, is this patch still necessary with the incorporation of display title support into Semantic MediaWiki? See http://sandbox.semantic-mediawiki.org/wiki/Display_Title_Test#tab=filtered. The filter labels in that example are correctly being replaced by the display title for the filtering pages.

Fannon:
@cicalese Oh, it looks like this solved the problem already. Then my patch is obsolete, as native DisplayTitle support is the much better solution for the problem!

cicalese:
Great!! I was advocating for a fix for datatables and am delighted that the fix also works for filtered, which we also use!

Fannon:
I fear that the problem isn't solved, sorry.

In the sandbox example you provided, the display title is used in the results itself, but not in the filter labels. There it's still using `Page1` instead of `Display Title Test Page 1`

Fannon:
I've just played around and found a way to fix this issue in a far more elegant way. I'll provide a new pull request when I'm done.

Fannon:
See #165

cicalese:
Just yesterday, the sandbox example was working correctly for filter labels. I know, because I upgraded one of my wikis to SMW 2.4 RC1 and could not reproduce the correct behavior locally, so I consulted the sandbox wiki to be sure that it was actually working correctly there. Perhaps a bug was introduced in some code that was updated since yesterday. @kghbln, is it possible to find out which extensions were upgraded since yesterday on the sandbox and from which version? Thanks!

Fannon:
In my new pull request, I've already fixed this problem. 

But as far as I understood the involved filtered format code, there is no possibility that it could have worked with display title regarding the filter checkboxes itself. It did definately make no distinction between id and label. Are you sure this part was working before? And if it did, it might have used the labels instead of the id's in general - which would break the format completely. 

This issusue (labels beeing used as ID's) is something I've also encountered with some Semantic Form Input types, like radio buttons.

Best,
Simon

cicalese:
Yes, the filter checkboxes were definitely showing up correctly with the semantic title being displayed rather than the page title. I did not check yesterday to make sure that they were correctly filtering the data when checked, but I did test that previously. I am stumped. Nonetheless, I would be happy to test your new patch as soon as I get the chance, but I may not have a chance to for a couple of days.

Fannon:
Very strange :) Thanks @cicalese !

kghbln:
> @kghbln, is it possible to find out which extensions were upgraded since yesterday on the sandbox and from which version? Thanks 

Yesterday sandbox did not upgraded at all since I deactivated the server. To find out which updates were done is however fairly easy. All changes are recorded with a lag of about 12 hours on [Wikiapiary](https://wikiapiary.com/wiki/Semantic_MediaWiki_-_Sandbox/Extensions), so for the night from the 20. to the 21. the [follwing extensions](https://wikiapiary.com/w/index.php?title=Semantic_MediaWiki_-_Sandbox%2FExtensions&type=revision&diff=2939240&oldid=2936209) were updated.

cicalese:
Thank you, @kghbln! I didn't think of using Wikiapiary for that!

kghbln:
> I didn't think of using Wikiapiary for that!

@cicalese You are welcome. This wiki really offers more info than one may expect at first. :)

Issue #145, opened 2016-01-24
timeseries format - 0 values are Excluded


jr5:
The "min" parameter seems somewhat culpable, but values of -1 will show on result if min is set to -2, 0 will never plot.   Harvest Amount is a property of has type::number

===List of eggs===

{{#ask:[[Has node type::Harvest]][[Has node start::+]][[Has node parent::{{PAGENAME}}]]
|?Harvest Type
|?Has node start=Has date 
|?Harvest Amount
|sort=Has node start
|order=asc
|default = No eggs to speak of
}}

=== Plot (Zero data is removed) ===

{{#ask:[[Has node type::Harvest]][[Has node start::+]][[Has node parent::{{PAGENAME}}]]
|?Has node start
|?Harvest Amount = Eggs
|format=timeseries    
|zoom=top
|group=property
|sort=Has node start
|order=asc
|headers=show  
|mainlabel=-
|layout=line
|height=400
|width=600
|charttitle=Egg Production
|min=-1
|charttext=Data was provided by our flock of hens. 

}}

jr5:
SRF_Timeseries.php line 94
if ( $timeStamp !== '' && ZZZZZZ &&  $rowSum >= $this->params['min'] ) {

ZZZZ Was $rowSum == true 
ZZZZ Resolved count($sum) > 0

mwjames:
> SRF_Timeseries.php line 94
if ( $timeStamp !== '' && ZZZZZZ && $rowSum >= $this->params['min'] ) {

Thanks.

Expected to be fixed with #149 using a slightly different assertion.

PR #149, opened 2016-02-17
Fix timeseries format with 0, refs #145

mwjames:
refs #145


mwjames:
http://sandbox.semantic-mediawiki.org/wiki/Issue/149_%28SRF%29

PR #150, opened 2016-02-17
Modify class identifier

mwjames:


mwjames:
> On line 53 there should be another replacement.

Good catch.

PR #156, opened 2016-05-13
Fix Special:Ask param descriptions for the excel format

s7eph4n:
* Use arrays instead of ParamDefinition objects
* Provide parameter descriptions

Set version to 2.4-alpha


JeroenDeDauw:
Thanks!

Issue #158, opened 2016-06-13
FilteredFormat: Since latest Chrome / Firefox Dev: Uncaught TypeError: fixedValues.split is not a function


Fannon:
Hello,

I've noticed this bug in FilteredFormats for a while happening in Chrome Dev and Firefox Dev browsers. The change leading to this bug seems to have recently landed in Chrome stable, as my main browser is having the same issue now, too.

Can someone else reproduce this, too?

## Steps to reproduce:
* Use a very recent browser (chrome dev, chrome latest, firefox nightly)
* Go to a page using the filtered-format result format with some filters active
* The filters are not displayed, as the JavaScript error stops the execution of the result formats script

![2016-06-13 12 24 44](https://cloud.githubusercontent.com/assets/470980/16004454/0697f974-3162-11e6-8271-bd9f004da2cb.png)

```
ext.srf.filtered.value-filter.js:157 Uncaught TypeError: fixedValues.split is not a function
init	@	ext.srf.filtered.value-filter.js:157
valueFilter	@	ext.srf.filtered.value-filter.js:305
(anonymous function)	@	ext.srf.filtered.js:117
each	@	load.php?debug=true&lang=en&modules=jquery%2Cmediawiki&only=scripts&skin=chameleon&version=20160610…:384
each	@	load.php?debug=true&lang=en&modules=jquery%2Cmediawiki&only=scripts&skin=chameleon&version=20160610…:136
attachFilter	@	ext.srf.filtered.js:103
$.fn.filtered	@	ext.srf.filtered.js:347
(anonymous function)	@	ext.srf.filtered.value-filter.js:317
(anonymous function)	@	ext.srf.filtered.value-filter.js:319
```

kghbln:
@mitchelln confirmed that C 51 starts to have the problem. Ping @s7eph4n as he was the original author of this cool format.

mitchelln:
Oh dear, this will be an absolute disaster. I have many sites running the filtered result format.

Happy to sponsor a fix for this.

Fannon:
I can confirm that this is fixed with pull request #160

Thanks, @s7eph4n !

kghbln:
So this will be in [SRF 2.4](https://github.com/SemanticMediaWiki/SemanticResultFormats/milestones/SRF%202.4)

PR #160, opened 2016-06-15
Filtered format: Fix Uncaught TypeError: fixedValues.split is not a function

s7eph4n:
Chrome 51+ and Firefox 48+ support the Array.prototype.values() method, which clashes with the 'values' parameter of the value filter. So we have to make sure, that when accessing that field we really get a string (i.e. something we set ourselves) and not a function (provided in the Array prototype).

See https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/Global_Objects/Array/values

### DIFF
--- a/RELEASE-NOTES.md
+++ b/RELEASE-NOTES.md
@@ -18,6 +18,8 @@ These are the release notes for the [Semantic Result Formats]
 * Fixed bug causing occasional exceptions in the calendar format (by Mark A. Hershberger)
 * Fixed bug in timeseries format that caused the value 0 to be excluded (by MWJames)
 * Fixed bug in the calendar parser functions (by James Montalvo)
+* Fixed bug in filtered format that broke the format on browsers supporting the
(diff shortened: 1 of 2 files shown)


mitchelln:
Hi Stephan.


Thanks for looking at this so quickly. I'll test the fix.


All the best

Neill.

________________________________

kghbln:
Cool! Refs https://github.com/SemanticMediaWiki/SemanticResultFormats/issues/158

mitchelln:
Hi.


I can confirm it as fixed. Great job [😊]
<https://www.wiltshireparentcarercouncil.co.uk/wpcc_dev/index.php/Children_Calendar>

https://www.wiltshireparentcarercouncil.co.uk/wpcc_dev/index.php/Children_Calendar


Many thanks

Neill.

________________________________

JeroenDeDauw:
Thanks

+1

Fannon:
Thanks! I can confirm that this fix works for me.

s7eph4n:
Thanks for testing, guys!

PR #165, opened 2016-06-23
Add displaytitle label support to filtered format, fix sorting order

Fannon:
This pull request is a proper version of the  #144 pull request.

It introduces support for displaytitle labels for the filters itself in the filtered format, given that at least SMW >= 2.4 is used. If SMW is of a lower version, this patch will fall back to use the ID as the label.

When testing, I found one catch: When `link=none` is used, no label can be detected, as it is extracted for the `formatted value` `<a>` tag. In that case there is no way for the result format to know what the label should have been.

This pull request also fixes a potential issue with the alphabetical ordering of the filters. Previously an object (dictionary) was used. Those do not guarantuee the order of their keys. Most browsers keep the order until a certain size, but after that they will shuffle it up.

Best,
Simon

### DIFF
--- a/formats/Filtered/libs/ext.srf.filtered.value-filter.js
+++ b/formats/Filtered/libs/ext.srf.filtered.value-filter.js
@@ -123,18 +123,31 @@
 			} );
 
 			var valueId; // just some valid value ID
+
+			/** Map of value => label distinct values */
+			var distinctValues = {};
+
+			/** Sorted array of distinct values */
+			var sortedDistinctValues = [];
+
 			if ( fixedValues == null ) {
 				// build filter values from available values in result set
 
 				// find distinct values and set visibility for all items that have
 				// some value for this printout
-				var distinctValues = [];
 
 				for ( valueId in values ) {
+
 					var printoutValues = values[valueId]['printouts'][target]['values'];
 
 					for (var j in printoutValues) {
-						distinctValues[ printoutValues[j] ] = true;
+						var formattedValue = values[valueId]['printouts'][target]['formatted values'][j];
+						var label = printoutValues[j] ;
+						// If the formatted Value is a link, it may contain a label (DISPLAYTITLE)
+						if (formattedValue.indexOf('<a') > -1) {
+							label = /<a.*>(.*?)<\/a>/.exec(formattedValue)[1];
+						}
+						distinctValues[ printoutValues[j] ] = label;
 					}
 
 					filtered.filtered( 'voteItemVisibility', {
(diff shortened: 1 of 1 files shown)


Fannon:
@cicalese Could you confirm that this adds proper displaytitle support?

formats/Filtered/libs/ext.srf.filtered.value-filter.js
```diff
@@ -254,11 +262,14 @@
...
 			// insert options (checkboxes and labels) and attach event handlers
 			// TODO: Do we need to wrap these in a form?
-			for ( var j in sortedDistinctValues ) {
+			for (var j = 0; j < sortedDistinctValues.length; j++) {
 				var option = $('<div class="filtered-value-option">');
+
+				// Try to get label, if not fall back to value id
+				var label = distinctValues[sortedDistinctValues[j]] || sortedDistinctValues[j];
+
 				var checkbox = $('<input type="checkbox" class="filtered-value-value" value="' + sortedDistinctValues[j] + '"  >');
```
JeroenDeDauw: 100 lines later and still the same function? [<o>](https://incabell.wordpress.com/2016/01/21/emoticon-suggestion/comment-page-1/)

JeroenDeDauw:
Thanks for contributing!

formats/Filtered/libs/ext.srf.filtered.value-filter.js
Fannon: I don't understand?

JeroenDeDauw: This function is huge. Very procedural programming style. I'm ranting about code quality here. (Which you can disregard if you want, I'm not objecting to this being merged.)

Fannon: Ah yes, that's the way the filtered format is written. My pull request doesn't do refactoring, but I've slimmed some things down, so overall it's propably less LOC than before.

Fannon: And I agree about the code quality. The whole file is just one function.

But procedural code isn't always that bad, at least its easy to understand and to reason about. You can also overengineer things to the point that they're even more of a maintainence burden than good old sphaghetti code ;) But enough philosophy here.

s7eph4n: Mea culpa re the coding style. I know it's a mess. It also has a number of design issues and if I had the time I would rewrite the whole thing from scratch.
@Fannon I'll have a look at this asap.

Fannon: Thanks @s7eph4n ! :+1:

Fannon:
Thanks for merging!

PR #173, opened 2016-09-01
Fixed datatables handling of empty printouts.

### LINKED ISSUE #172: Datatables format / empty printout not supported
* MediaWiki 1.25.3
* SMW 2.4.1
* SRF 2.3

Sometimes I want to move the subject of a query result to a different position in a table row. For the standard formats, this is done by using `mainlabel=1` and then a printout like `|?=Subject`. However, if you try and do this with datatables you get a Javascript error like this:
````
DataTables warning (table id = 'DataTables_Table_0'): Requested unknown parameter 'Subject' from the data source for row 0
````

Sandbox example to come hopefully - it's just being a little bit slow at the moment.

kriegfrj:
Fixes #172, regression test included.

### DIFF
--- a/formats/datatables/resources/ext.srf.formats.datatables.js
+++ b/formats/datatables/resources/ext.srf.formats.datatables.js
@@ -144,11 +144,10 @@
 				}
 
 				// Transform results into a specific aaData format
-				function getResults( parameters, results ){
+				function getResults( parameters, results, printreqs ){
 
 					var aaData = [],
 						i = 0;
-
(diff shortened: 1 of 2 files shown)


JeroenDeDauw:
Thanks a lot for the fix!

JeroenDeDauw:
Ping @mwjames

mwjames:
I haven't tested it but to avoid code rot (we always can revert), I'd rather see this deployed.

PR #182, opened 2016-10-17
Excel format: Throw error if phpExcel is missing (+ minor rework)

s7eph4n:


kghbln:
@JeroenDeDauw @s7eph4n Is this cool?

PR #202, opened 2017-03-16
Fixes #135

### LINKED ISSUE #135: Thousands separator in math formats (English language MW)
It seems that it is not possible to get a result from the **math** [0] formats without the thousands separator. With `{{#ask:...` or `{{#show:...` you can specify if the thousands separator should be in the result or not but this is not transferred to any of the math formats. There are no parameters documented [0]. This means that the results of these formats can not directly be used in the `{{#expr:...` parser function. If I could I would disable the thousands separator all together but that is not possible [1]. See example [2].

[0] http://www.semantic-mediawiki.org/wiki/Help:Sum_format
[1] http://www.semantic-mediawiki.org/wiki/Help:Type_Number#Numbers_in_other_languages
[2] http://sandbox.semantic-mediawiki.org/wiki/Thousands_separator

SchmidDev:
Fixes #135

### DIFF
--- a/formats/math/SRF_Math.php
+++ b/formats/math/SRF_Math.php
@@ -4,7 +4,7 @@
  * Various mathematical functions - sum, product, average, min and max.
  *
  * @licence GNU GPL v3+
- * 
+ *
  * @author Jeroen De Dauw < jeroendedauw@gmail.com >
  * @author Yaron Koren
  * @author Nathan Yergler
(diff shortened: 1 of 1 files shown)


kghbln:
Thanks for your pull request. I have requested reviews now.

gesinn-it:
@JeroenDeDauw can this PR be merged?

mwjames:
Please be so kind and verify the change by creating an example at [0] and link it to this issue. Thanks.

[0] https://sandbox.semantic-mediawiki.org

jongfeli:
Thanks for this. 

I tested this on [0]. It seems that the decimal symbol changed to "**,**" on the SMW Sandbox. Now the property values must have a comma as a decimal separator.

It is better to look at the test page but here is an attempt to write down the reult:
- A plain Ask result (no format) the thousands sep is a **comma** and the decimal sep is a **point**. 
- A plain Ask result (no format) without the thousands separator shows the decimal sep with a **comma**. 
- When using **min**, **max** or **sum** the decimal separator is the same with or without the thousands separator but it is a **point** and not a **comma** as the property value.

Stripping the thousands separator in the math formats works just fine but it seems that the handling of the separator types is not correct in all cases.

[0] https://sandbox.semantic-mediawiki.org/wiki/Thousands_separator

gesinn-it:
@jongfeli Thanks for creating the test cases in the sandbox. You are right. Having a look into this...

gesinn-it:
@mwjames @JeroenDeDauw  @kghbln : In [1] I can not find any specification, how the decimal separator is handled in plain printouts. As shown in @jongfeli test page [2], it seems to be always ',' independent of user language. If this is the case, should we change the Math result format to also always use ',' as decimal separator?

[1] https://www.semantic-mediawiki.org/wiki/Help:Type_Number
[2] https://sandbox.semantic-mediawiki.org/wiki/Thousands_separator

gesinn-it:
I vote for '.' as decimal separator to make raw results directly consumable by ParserFunctions [1].

[1] https://www.mediawiki.org/wiki/Help:Extension:ParserFunctions#.23expr

PR #209, opened 2017-04-27
Update datatables format to handle format param on number and quantity

ljonka:


kghbln:
Thanks a lot for the pull request! I requested a review.

mwjames:
Thanks.

I cannot promise that I have time this weekend for a review but some details or examples that explain this change would be welcome.

ljonka:
Hello james, this change add format handling to datatables for number and quantity to show fractional digits as requested from user in #ask question, eg ?numberfield#-p3, currently this is ignored.

kghbln:
@mwjames Coundn't this be also an improvement for class datatable in SMW 3.0.0.

mwjames:
> Coundn't this be also an improvement for class datatable in SMW 3.0.0.

Not really because the values are already formatted in accordance with the printer preferences on the output format and represented as "pure" HTML by the time the `datatable` gets to transform the HTML.

The above case as well as the `datatables` format work a bit different from that it uses an object (DI) to modify the representation of a value.

I'm still a bit unsure about the use of `Intl.NumberFormat( 'latn', { minimumFractionDigits: decNum[2], maximumFractionDigits: decNum[2] } )` and its implications that's why I haven't merged it yet.

kghbln:
> Not really because the values are already formatted in accordance with the printer preferences on the output format and represented as "pure" HTML by the time the datatable gets to transform the HTML.

Ah, yeah, that makes sense.

ljonka:
according to https://www.mediawiki.org/wiki/Manual:Installation_requirements/en#PHP installation of php-intl is explained.

mwjames:
> according to https://www.mediawiki.org/wiki/Manual:Installation_requirements/en#PHP installation of php-intl is explained. SMW uses another formater: 

I'm aware of this but my issue is with `Intl.NumberFormat` in JS.

mwjames:
I haven't forgotten about this, it is just that SMW-core (or better the issues open in MW-core) keep me from doing other things during my volunteer support window.

ljonka:
see hallowelt fork for working version.

PR #212, opened 2017-05-13
Prepare SRF when SMW is loaded via extension.json

mwjames:
- Move checks to `$wgExtensionFunctions` to ensure SMW is loaded before it is checked especially when SMW `extension.json` is involved 
- Streamline `SemanticResultFormats` to work as any other extension in https://github.com/SemanticMediaWiki
- Renamed `SemanticResultFormats.settings.php` to `DefaultSettings.php`


mwjames:
@kghbln FYI; Everything should work as before but it would have definitely not worked after #1732 got merged.

kghbln:
> Everything should work as before

Affirmative. No disruptions whatsoever on sandbox since this was merged.

PR #225, opened 2017-06-07
Add option to use DEFAULTSORT in Filtered format

matthew-a-thompson:
When using the Filtered result format, a field specified as a value filter when using can have its ordering specified explicitly using the "value filter values" parameter, or it can use a default behavior of alphabetically sorting based on the value of that field. However, neither of these behaviors are particularly useful when the filter is on a field where the values are page names (for example, something like Person:123): normal alphabetization would put them in a strange order (where Person:123 precedes Person:2, for instance), and when an explicit order listing (which would have to be generated by some other function) is used, the same value is used both for the checkbox label as for the sought value in the field (so a checkbox that was labeled "Thompson, Matthew" would filter based on that string, which would match no results, as the value in the result is e.g. "Person:123").

A desirable behavior in this case would be to use the page's DEFAULTSORT attribute to establish the ordering. To this end, I have added an optional parameter "value filter pagedefaultsort" to the Filtered result format. When this parameter is set (it does not matter to which value the parameter is set), additional sort information is generated based on the DEFAULTSORT attribute of the pages that are found when using the field's value as a page title. When this information is present in a field that is being used to create filter checkboxes, the sort is applied, and the checkboxes appear in the order specified by DEFAULTSORT.


kghbln:
Thank you for this pull request. I will now add review requests.

matthew-a-thompson:
Thank you for the detailed feedback, @s7eph4n . I'll try to get my changes ready so they can be merged soon after your update. I have two questions for you.
1. In regards to the requested change to not access the database, I was unable to discover an alternative method of retrieving the DEFAULTSORT property of a page without using the `PageProps::getProperties` method that you mentioned. Since supporting versions prior to 1.27 is a requirement, would you be willing to accept using the `PageProps` class if it exists, and querying the database directly as a fallback?
2. Right now, the special sorting behavior only occurs when the new "value filter pagedefaultsort" option is set. However, it might make more sense to make the use of the DEFAULTSORT property the default behavior in cases where the value of a field is a page that has a DEFAULTSORT property set on it. When this property is not available (or the value in question is not a page), the existing behavior would be used. Would you be willing to accept this change as a new default behavior, rather than as a special behavior enabled by a new parameter?

s7eph4n:
Yeah, it appears that they did the DB access directly prior to 1.27, so a feature switch is probably the way to go.

The idea of using DEFAULTSORT as default behavior is really good, I think. In fact I think we should just drop the parameter and make this the standard. If somebody specified a DEFAULTSORT then they probably meant it to be used.

formats/Filtered/SRF_Filtered_Item.php
```diff
@@ -51,6 +51,7 @@ public function getArrayRepresentation() {
 
 		$printouts = [];
 		$isFirstColumn = true;
+		$dbr = wfGetDB( DB_SLAVE ); // might need this soon
```
mwjames: I saw this and it made my cry.

mwjames:
> Yeah, it appears that they did the DB access directly prior to 1.27, so a feature switch is probably the way to go.

I'm thinking that maybe we should kick 1.23 from the list of supported releases after it has been EOL since 05.2017 [0] and go for 1.27 as min. SMW 2.5 still has support for 1.23 but that was released while 1.23 was still an option (when only by a margin) and has a broader user base, so the situation was (is) a bit different.

> I think. In fact I think we should just drop the parameter and make this the standard. If somebody specified a DEFAULTSORT

:+1:
Using DEFAULTSORT is a deliberate act and should be seen as an intentional decision (as we do in SMW-core) therefore supporting it as a default mode seems sensible.

[0] https://www.mediawiki.org/wiki/MediaWiki_1.23

kghbln:
> I'm thinking that maybe we should kick 1.23 from the list of supported releases after it has been EOL since 05.2017 [0] and go for 1.27 as min.

In this case we have to officially [break with old traditions](https://www.semantic-mediawiki.org/wiki/Extension:Semantic_Result_Formats#Version). Actually we already did, so :+1:

s7eph4n:
Although I think that this tradition would be worth keeping. Makes life much easier for users. (And developers who will have to deal with users assumptions of compatibility and resulting requests for support. :wink: )

kghbln:
> Although I think that this tradition would be worth keeping. Makes life much easier for users.

At the time we did not have Composer. Nowadays it is the task of Composer to take care of this - based on developer input though.

s7eph4n:
Well, we also have tar balls for SMW and, much as I dislike dealing with them, I think we should also have them for SRF.

mwjames:
> Although I think that this tradition would be worth keeping. Makes life much easier for users. 

I don't mind, if you think it is worth the effort, keep it.

kghbln:
> Well, we also have tar balls for SMW and, much as I dislike dealing with them, I think we should also have them for SRF.

Hmm, so lets stick to it.

> I don't mind, if you think it is worth the effort, keep it.

At this point it should be easy:

I figured that SMW 2.5.0 still works with PHP 5.3 but it does not. so https://github.com/SemanticMediaWiki/SemanticResultFormats/commit/fa4c91b50794dce8d90c6696149538fde2f29831 is still ok and does not need reverts or so.

matthew-a-thompson:
I see that the changes to the Filtered format have been pulled in, so I'll get to work on updating my own changes for this pull request to incorporate the usage of DEFAULTSORT.

s7eph4n:
Thanks for the patch!

kghbln:
> Thanks for the patch!

Is it possible to back-port to 2.x, i.e. does it run with 1.23.x

kghbln:
> @s7eph4n Is it possible to back-port to 2.x, i.e. does it run with 1.23.x

Poke.

s7eph4n:
Sorry, I'm not at home right now. Poke me next Tuesday if you have not heard back.

kghbln:
> Sorry, I'm not at home right now. Poke me next Tuesday if you have not heard back.

Thanks for your info. Now worries. That's early enough for the upcoming release.

Issue #190, opened 2017-02-22
2.5.0 release


kghbln:
Tracking issue for the 2.5.0 release.
- [Release notes](https://github.com/SemanticMediaWiki/SemanticResultFormats/blob/master/RELEASE-NOTES.md)
- [Changes since 2.4.2](https://github.com/SemanticMediaWiki/SemanticResultFormats/compare/2.4.2...master)
## 
- [x] Finalize RELEASE NOTES also by clearing the ["releasenotes"](https://github.com/SemanticMediaWiki/SemanticResultFormats/issues?q=label%3Areleasenotes+is%3Aclosed) label
- [x] Update version number `SemanticResultFormats.php` (tag is 2.5.0 NOT 2.5)
- [x] Update version number `README.md`
- [x] Update `INSTALL.md` (compat)
- [x] Create tag
- [x] Create 2.5.x branch
- [x] Announce via mail
- [x] Do a release tweet on Twitter
- [x] Run `composer update` on smw.o

kghbln:
@JeroenDeDauw @mwjames Do you think it is time to prep for this and drop compat with MW < 1.23 on master.

JeroenDeDauw:
+1... kinda odd master of this thing still supports < 1.23, so ancient now.

s7eph4n:
SMW 2.5 requires `"php": ">=5.5.0"`. Should we raise the required PHP version for SRF as well?

kghbln:
> SMW 2.5 requires "php": ">=5.5.0". Should we raise the required PHP version for SRF as well?

I think this may be a very good idea.

JeroenDeDauw:
Oh wow it still supports 5.3 >_>

kghbln:
In the light of [this discussion](https://github.com/SemanticMediaWiki/SemanticResultFormats/pull/225#issuecomment-307546585): Let's make the 2.5.0 today and move on with future stuff. Objections @mwjames @s7eph4n ?

s7eph4n:
Yes, your honor, I'd have some, if I may. :)
If possible I would like to see the [reworked filtered format](/s7eph4n/SemanticResultFormats/tree/filtered) in 2.5. I am working to add some integration tests today and will then open the pull request.

kghbln:
> If possible I would like to see the reworked filtered format in 2.5. I am working to add some integration tests today and will then open the pull request.

[...]

PR #230, opened 2017-06-12
Cleanup, polish install instructions, add to release notes

s7eph4n:
* remove temporary package installations from travis build script
* do not load filtered by default (it loads map data from external page)
* polish INSTALL.MD
* add changes to RELEASE-NOTES.md

This PR includes:
- [x] CI build passed

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -169,7 +169,7 @@ Add the following line to the end of the "require" section in your "composer.loc
    * Remark 2: If you do not have a `composer.local.json` file (MediaWiki <1.25),
      use `composer.json` instead.
 
-   * Remark 3: If you do not have a `composer.json` file (MediaWiki <1.23.5),
+   * Remark 3: If you do not have a `composer.json` file (MediaWiki <1.24),
      copy `composer.json.example` to `composer.json` first.
 
 #### Step 4
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -139,38 +139,45 @@ minimum requirements are indicated in bold.
 
 ### Composer Installation
 
-The recommended way to install Semantic Result Formats is with [Composer](http://getcomposer.org)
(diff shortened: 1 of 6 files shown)


INSTALL.md
```diff
@@ -139,38 +139,45 @@ minimum requirements are indicated in bold.
...
+   * Remark 1: Remember to add a comma to the end of the preceding line in this 
+     section.
 
-You should already have a "composer.local.json" file due to installing the required extension
-Semantic MediaWiki. If not you should install Semantic MediaWiki first before you continue.
+   * Remark 2: If you do not have a `composer.local.json` file (MediaWiki <1.25),
+     use `composer.json` instead.
 
-When this is done run in your shell:
+   * Remark 3: If you do not have a `composer.json` file (MediaWiki <1.23.5),
```
kghbln: This may go since this release is not supposed to support MW <1.23 any longer

kghbln:
@s7eph4n Thanks for fluff. :)

INSTALL.md
s7eph4n: But 1.23 .0 < 1.23.5. ;-)

kghbln: What a mess MW created. I figured that 1.23.x already used `composer.json` itself. This acutually started with 1.24.0. So shouldn't it be 1.23.17 ;)

s7eph4n: Not sure. I just copied over the remarks from Chameleon without checking them again. I think what happened is that at the time of writing 1.23.5 was the latest on the 1.23 branch. We should probably just put <1.24.

kghbln: Probably maybe. Or 1.23.x to make sure what people do not think of 1.22. But this is nitpicking. :)

s7eph4n: Done. In any case, it should only be in for a few months.

PR #232, opened 2017-06-13
Make 2.5.0 release

kghbln:

- Provides the 2.5.0 release
- Updated RELEASE-NOTES
- Updated INSTALL

This PR includes:
- [ ] CI build passed

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -5,6 +5,7 @@
 	"keywords": [
 		"SRF",
 		"Semantic Result Formats",
+		"Result Formats",
 		"SMW",
 		"Semantic MediaWiki",
 		"Wiki",
@@ -46,7 +47,7 @@
 	"require": {
 		"php": ">=5.5.0",
 		"composer/installers": "1.*,>=1.0.1",
-		"mediawiki/semantic-media-wiki": "~2.1|~3.0",
+		"mediawiki/semantic-media-wiki": "~2.1",
 		"nicmart/tree": "^0.2.7",
 		"symfony/css-selector": "^3.3"
 	},
--- a/composer.json
+++ b/composer.json
@@ -47,7 +47,7 @@
 	"require": {
 		"php": ">=5.5.0",
 		"composer/installers": "1.*,>=1.0.1",
-		"mediawiki/semantic-media-wiki": "~2.1",
+		"mediawiki/semantic-media-wiki": "~2.1|~3.0",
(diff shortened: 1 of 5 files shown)


composer.json
```diff
@@ -46,7 +47,7 @@
 	"require": {
 		"php": ">=5.5.0",
 		"composer/installers": "1.*,>=1.0.1",
-		"mediawiki/semantic-media-wiki": "~2.1|~3.0",
+		"mediawiki/semantic-media-wiki": "~2.1",
```
s7eph4n: If you do this, there may be no stable SRF version available to go with SMW 3.0. At least as long as it takes SRF to catch up, which was several months for 2.5.

kghbln: I was going to add this to master after release. This time we should be more energetic I guess. Will add.

s7eph4n: Agree to being more energetic. Or just release what's available (as long as there are no catastrophic bugs).

kghbln: Still may be changed on the 2.5.x branch later on

mwjames:
> If you do this, there may be no stable SRF version available to go with SMW
> 3.0. At least as long as it takes SRF to catch up, which was several months
> for 2.5.

I don't know what we will break, I'd rather avoid users pull in
something by accident. Doing a simple 2.5.1 (or what comes hereafter)
can be done at any time.

PS: I don't really like the tarball stuff, it cause a lot of work with
no real benefit. If could be automated [0], I wouldn't mind otherwise
it is surely a nuisances.

[0] https://docs.travis-ci.com/user/deployment/releases/

Cheers

kghbln:
We will definitely have a 2.5.1 and if it is only for translation updates so this may be changed on the 2.5.x branch without worries I guess.

s7eph4n:
Well, true regarding releasing a minor version. Although we did just point out that a release is somewhat time-consuming. :)

And definitely agreed on the automatic building of stuff. Ideally the whole release process (incl. messaging) should be automatic.

mwjames:
> And definitely agreed on the automatic building of stuff. Ideally the whole
> release process (incl. messaging) should be automatic.

Patch releases should not need an announcement, yet for SMW the
situation is slightly different because as of now I back-ported some
features to the 2.5.x branch otherwise it would take month before they
become available and I'm not going to open the can for a full featured
2.6/2.7 release.

I'm not expecting that to happen for SRF.

PR #233, opened 2017-06-13
Start 3.0.0 development version

kghbln:

- Starts the development of SRF 3.0.0 on master since 2.5.x is now on its own branch

This PR includes:
- [ ] CI build passed

### DIFF
--- a/SemanticResultFormats.php
+++ b/SemanticResultFormats.php
@@ -51,7 +51,7 @@ class SemanticResultFormats {
 	 */
 	public static function initExtension() {
 
-		define( 'SRF_VERSION', '2.5.0' );
+		define( 'SRF_VERSION', '3.0.0-alpha' );
 
 		// Register the extension
 		$GLOBALS['wgExtensionCredits']['semantic'][] = [
--- a/composer.json
+++ b/composer.json
@@ -72,7 +72,7 @@
 	},
 	"extra": {
 		"branch-alias": {
-			"dev-master": "2.5.x-dev"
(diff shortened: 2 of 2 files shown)


s7eph4n:
@kghbln, I just added a checkbox to the pull request template to update the release notes with every PR. I think this should help to minimize the effort to release new versions. It means however, that there should always be a section in the release notes for upcoming versions, where updates can be put.
What do you think? :wink:

kghbln:
@s7eph4n 

> I just added a checkbox to the pull request template to update the release notes with every PR. I think this should help to minimize the effort to release new versions.

Great, good idear

> It means however, that there should always be a section in the release notes for upcoming versions, where updates can be put.

You mean link this: https://github.com/SemanticMediaWiki/SemanticResultFormats/commit/ef955e0cb19baecfeb161842b088d1cc7edef19b :) I think in the past the first one with something started this but this indeed provides more guidance.

mwjames:
https://github.com/SemanticMediaWiki/SemanticResultFormats/commit/ef955e0cb19baecfeb161842b088d1cc7edef19b
> :) I think in the past the first one with something started this but this
> indeed provides more guidance.

If possible, I still like to see the back-ported sha1 otherwise if you
not link the #... number as reference it is difficult to find the
exact port.

kghbln:
> If possible, I still like to see the back-ported sha1 otherwise if you not link the #... number as reference it is difficult to find the exact port.

Yeah, however the items are already in the list which will make it easier to identify and see what happened.

Issue #248, opened 2017-06-22
Sum format returns a space for Kiloseparator


camponez:
### Setup
- MW version: 1.28.2
- DB (MySQL etc.): 10.2.6-MariaDB-10.2.6+maria~jessie
- PHP version:7.0.16 (fpm-fcgi)
- SMW version: 2.5.2
- SRF version: 2.5.0
- Browsers and versions (if applicable): ALL

### Issue

Sum format returns space for Kiloseparator.

**Steps to reproduce**

Use {{#ask: PARAMETER | format=sum}}

**Expected result:** ...

Either: with the kilo separator defined by the locale (in my case 'a dot')
Or: a standard number format (no kiloseparator and 'a dot' for decimal)

**Observed result:** ...

Kilo separator is a space.

mwjames:
> **Observed result:** ...
>
> Kilo separator is a space.

@jaideraf Could you comment here on the pr-br matter.

As defined in [0], I would expect the decimal as `,` (comma) and kilo
being a ` ` (space) therefore a number like `3,000.2` in English is
displayed in "português do Brasil" as `3 000,2`.

[0] https://github.com/SemanticMediaWiki/SemanticMediaWiki/commit/e7dea7f0349799fdfee307e44c8f8c184a945e7b

camponez:
@mwjames 
1 - I think it should be `.` for pt-br
2 - I don't see that line on the SW I have installed.

camponez:
When using `table` format all numbers are correctly formatted.

mwjames:
> When using `table` format all numbers are correctly formatted.

I'm confused now, or I misinterpreted the issue description.

I would expect (given the previously outlined rules of comma and
space) that a result for the number 3000.2 (ISO format) is displayed
both in `sum` and `table` format (for user language pt-br) as:

3 000,2

camponez:
>I would expect (given the previously outlined rules of comma and space) that a result for the number 3000.2 (ISO format) is displayed both in `sum` and `table` format (for user language pt-br) as:
> 3 000,2

That's exactly the issue. It only shows `space` for kiloseparator when I use `sum` format. For `table` it works correctly.

Sum format: `3 000,02`
Table format: `3.000,02`

camponez:
Here it works: https://is.gd/bK22hl
Here it doesn't: https://is.gd/YOoXJs

mwjames:
>
> Sum format: `3 000,02`
[...]

PR #250, opened 2017-07-01
Math format to use getLocalizedFormattedNumber, refs 248

mwjames:
This PR is made in reference to: #248

- Use `NumberValue::getLocalizedFormattedNumber` to get the localized format according to settings available
- Cannot be back-ported to SRF 2.5.x (as that would allow SMW 2.1) [0] because the change requires at least SMW 2.4.x

[0] https://github.com/SemanticMediaWiki/SemanticResultFormats/blob/2.5.x/composer.json#L50

This PR includes:
- [x] CI build passed

### DIFF
--- a/formats/math/SRF_Math.php
+++ b/formats/math/SRF_Math.php
@@ -44,8 +44,8 @@ class SRFMath extends SMWResultPrinter {
 
 		// if raw-format ("-") than skip formatNum()
 		if ( $outputformat != "-" ) {
-			global $wgLang;
-			$number = $wgLang->formatNum( $number );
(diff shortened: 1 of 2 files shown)


mwjames:
It currently fails and requires SMW 2.5.3 where related messages for pt-br were updated in order for the integration test to pass.

kghbln:
@mwjames So this could be backported to 2.x. Will do SRF 2.5.1 the same day as SMW 2.5.3 if that's ok.

kghbln:
> Cannot be back-ported to SRF 2.5.x (as that would allow SMW 2.1) [0] because the change requires at least SMW 2.4.x

@mwjames Never mind my previous post.

Issue #253, opened 2017-07-06
Tagcloud format: the align attribute on the div element is obsolete.


jaideraf:
### Setup

- MW version: 1.28.2
- DB (MySQL): 5.6.36-82.0
- PHP version: 7.0.20 
- SMW version: 2.5.2
- SRF version: 3.0.0-alpha

### Issue

Just a minor issue, from the W3C Validator, showing that `tagcloud` format is using the `align` attribute on div elements.

**Steps to reproduce**

[See the W3C Validator message](https://validator.w3.org/nu/?doc=https%3A%2F%2Fwww.semantic-mediawiki.org%2Fwiki%2FDemo%3ATagcloud%2FTagcloud_using_the_sphere_widget): 

```
Error: The align attribute on the div element is obsolete. Use CSS instead.
<div class="srf-tagcloud-sphere" data-version="0.4.1" align="justify">
```

**Expected result:**
Maybe just removing it (it did not changed the presentation to me):
```
<div class="srf-tagcloud-sphere" data-version="0.4.1">
```

jaideraf:
Aditionally Chrome says:

```
This page is using the deprecated ResourceLoader module "jquery.ui.position".
This page is using the deprecated ResourceLoader module "jquery.ui.widget".
This page is using the deprecated ResourceLoader module "jquery.ui.core". 
Please use "mediawiki.ui.button" or "oojs-ui" instead.
```
As the author, maybe @JeroenDeDauw is interested in this.

kghbln:
Thanks for reporting.

kghbln:
I am not sure what the point of `align="justify"` was anyways. That was supposed to stretch the content to the with of the parent container so every word is displayed in equal width.

@jaideraf  It this really something one would want for a tagcloud? I myself do not. Besides, I think you do not see a difference before and after since `align="justify"` is no longer supported by the browser anyways.

If someone wants justify her or his data she or he could add `text-align: justify;` to the "srf-tagcloud-sphere" class.

jaideraf:
> @jaideraf It this really something one would want for a tagcloud? I myself do not. 

No. Me, too. If nothing changes, with or without `align="justify"`, it's better without it.

kghbln:
@jaideraf Thanks for your feedback confirming my view. The fix will ship with SRF 2.5.1 most probably next week.

PR #254, opened 2017-07-06
Update TagCloud.php

kghbln:
This PR is made in reference to: #253 

- Remove obsolete "div" element `align="justify"`

This PR includes:
- [x] Update of RELEASE-NOTES.md
- [x] CI build passed

### DIFF
--- a/formats/tagcloud/TagCloud.php
+++ b/formats/tagcloud/TagCloud.php
@@ -314,8 +314,7 @@ class TagCloud extends ResultPrinter {
 		// General placeholder
 		$attribs =  [
 			'class'  => 'srf-tagcloud' . $class,
-			'data-version' => '0.4.1',
-			'align'  => 'justify'
+			'data-version' => '0.4.1'
 		];
 
 		return Html::rawElement( 'div', $attribs, $processing . $htmlSTags );


kghbln:
Note self: Needs to be backported to 2.5.x

s7eph4n:
Needs testing. 😛 

No really, it has never been easier. Just specify a test setup and then use a CSS selector that expects not to see a certain attribute.

See [0] for an example test, [1], [2], [3] for some relevant CSS docu.

[0] https://github.com/SemanticMediaWiki/SemanticResultFormats/blob/6077519/tests/phpunit/Integration/JSONScript/TestCases/Filtered-01.json
[1] https://stackoverflow.com/questions/1533444/css-selector-to-match-an-element-without-attribute-x#1533485
[2] https://developer.mozilla.org/en-US/docs/Web/CSS/:not
[3] https://developer.mozilla.org/en-US/docs/Web/CSS/Attribute_selectors

kghbln:
> No really

@s7eph4n This is part of a bigger issue. So far this result format is completely untested which needs to be changed. Testing for something not present should be part of it. I would like to move this to another issue rather than having this as holdup. In the end we are removing "functionality" ;)

JeroenDeDauw:
@s7eph4n I'm not sure what you would test here. You can test that the CSS attribute is there or not, though that does not tell you anything about if it is doing the job it is supposed to be doing or not. So while tests for this format and stuff in general are cool, I don't think it makes sense to write any for this PR.

JeroenDeDauw:
@kghbln not sure it's worth spending time on backporting a change like this one

s7eph4n:
@JeroenDeDauw Well, the issue was that the `align` attribute is obsolete and should be removed. It was not that the output of the gallery format looks bad.
So I would test that that issue is indeed fixed, i.e. the attribute not present anymore. Although I agree that testing that something is not there anymore is really a special case test.

PR #259, opened 2017-07-09
Transfer html validator

s7eph4n:
This PR is made in reference to: [SMW#2531](https://github.com/SemanticMediaWiki/SemanticMediaWiki/issues/2531)

This PR removes ParserHtmlTestCaseProcessor et al. It is part of SMW now.

This PR includes:
- [x] CI build passed


s7eph4n:
Good point.
Should we set the requirement to ~3.0 then? But that would possibly lead to problems while SMW 3.0 is still in development, wouldn't it? At the least users would have to explicitly require SMW 3.0-devel on top level to work with SRF 3.0-devel.

mwjames:
> And I'm guessing this class is not in SMW 2.1

It doesn't really matter, this change is concerned with the testing
environment and not with some code path used in production.

The SMW part can be back-ported to 2.5.x and should available to SRF
if tests against 2.5.x@dev or 3.x.

JeroenDeDauw:
> It doesn't really matter, this change is concerned with the testing environment and not with some code path used in production.

That makes a bit of a difference, though it still matters, because it prevents you from testing compatibility with the oldest version of SMW you claim to support. Apparently the test are currently only run with latest SMW, which is not ideal.

s7eph4n:
According to https://www.semantic-mediawiki.org/wiki/Extension:Semantic_Result_Formats#Version the version requirement should really be `"mediawiki/semantic-media-wiki": "~2.5|~3.0"`.

What about branching of a 2.5.x branch parallel to SMW 2.5.x and then going forward on master requiring `"mediawiki/semantic-media-wiki": "dev-2.5.x|dev-master"` while SMW 3.0 is in delevopment, later changing to `"mediawiki/semantic-media-wiki": "~3.0"`?

kghbln:
> What about branching off a 2.5.x branch parallel to SMW 2.5.x

This branch already exists and e.g. was used to release 2.5.1 this afternoon.

JeroenDeDauw:
"dev-2.5.x|dev-master"

would better be

    "2.5@dev|3.0@dev"

Though perhaps you can just drop the 2.5 part already. The later version range spec includes 2.5.0 and 2.5.1 (stable releases), which is not the case if you just point to a dev branch.

s7eph4n:
Hmm, according to https://getcomposer.org/doc/04-schema.md#package-links stability flags only work at root-level. So just '~3.0'?

JeroenDeDauw:
Euhm... this is at the root level?

s7eph4n:
Maybe I don't get it, but as far as I understand it SRF when installed using composer becomes a sub-component of that MW installation.

JeroenDeDauw:
Uhm... I'm not up to speed with the non-root level stuff, but I can say that you can use stability modifiers here as I suggested

s7eph4n:
Ok, will do. Worst that can happen is they have no effect. :)

s7eph4n:
@JeroenDeDauw Thanks for merging #262. Mind having a look at this as well?

s7eph4n:
Thanks! :)

PR #277, opened 2017-07-23
Enable test coverage reports

s7eph4n:
This PR addresses enables test coverage reporting.

Before this is merged somebody should enable external code coverage on Scrutinizer. See https://scrutinizer-ci.com/docs/tools/external-code-coverage/

This PR includes:
- [ ] CI build passed

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,7 +5,7 @@ matrix:
   include:
     - env: DB=mysql; MW=master; SMW=~3.0@dev
       php: 5.6
-    - env: DB=mysql; MW=1.28.2; SMW=~3.0@dev
+    - env: DB=mysql; MW=1.28.2; SMW=~3.0@dev; TYPE=coverage
       php: 7.0
     - env: DB=sqlite; MW=1.28.2; SMW=~3.0@dev
       php: 7.0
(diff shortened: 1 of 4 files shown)


JeroenDeDauw:
You know you can use branches on this repo instead of your own right? You do have write access

s7eph4n:
Yes, I know. In principle. It never occurred to me though. :D 
Btw, "somebody should enable external code coverage on Scrutinizer". And that I really can't do. :P

JeroenDeDauw:
Yes you can, you just need to update https://github.com/SemanticMediaWiki/SemanticResultFormats/blob/master/.scrutinizer.yml

This needs to be added: https://github.com/wmde/Euro/blob/master/.scrutinizer.yml#L5

JeroenDeDauw:
Might need to specify custom timeout limit to avoid timeouts if the tests here are not that fast https://github.com/wmde/Diff/blob/master/.scrutinizer.yml#L5-L6

s7eph4n:
Abyssmal, as expected: [Code Coverage](https://scrutinizer-ci.com/g/SemanticMediaWiki/SemanticResultFormats/?branch=master)
And that already includes integration tests.

JeroenDeDauw:
Better than most MW extensions :) Also, this number does not include the JS, which this extension has a lot of, so it is actually lower.

PR #200, opened 2017-03-08
eventcalendar format / added clicktarget parameter

FelixAba:
to allow e.g. opening a form in edit mode when clicking on an calendar day. See commit messages for details.

PR #284, opened 2017-07-31
Fixed calendar rendering in IE

kwji:
Fixed syntax error in onDayClick function declaration.

When using the eventCalendar result format IE users are presented with an infinite loading spinner. 

The reason for this is that the JS function definition for the function "onDayClick" is syntactically incorrect and causes IE to choke. Specifically, it currently looks like:

`onDayClick( date, data, clickPopup )`

when it should instead be:

`onDayClick: function( date, data, clickPopup )`

### DIFF
--- a/formats/calendar/resources/ext.srf.formats.eventcalendar.js
+++ b/formats/calendar/resources/ext.srf.formats.eventcalendar.js
@@ -418,7 +418,7 @@
 		/**
 		 * Handles redirect to a clicktarget URL.  
 		 */
-		onDayClick( date, data, clickPopup ){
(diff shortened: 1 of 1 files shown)


s7eph4n:
Thanks for the pull request! It seems a simple enough fix and looks good to me, but the format was originally written by @mwjames , so I'll leave it to him to merge.

mwjames:
> was originally written by @mwjames , so I'll leave it to him to merge.

Issue introduced in #200 by @FelixAba.

kghbln:
Back-ported to 2.5 with 46bbbc778cece4b8a3dc5506f38d1197ec105c14.

PR #288, opened 2017-08-13
Filtered: Fixes

s7eph4n:
- Commit npm's package-lock.json
- New leaflet version
- Fix Number filter when only one value is available
- Fix for PHP's `next` not returning all values of an array (threading issue?)
- Error message when using Map view with string-type coordinates and data-values/geo missing

This PR includes:
- [ ] CI build passed


formats/filtered/src/Filtered.php
```diff
@@ -164,6 +165,7 @@ protected function getResultText( SMWQueryResult $res, $outputmode ) {
 		$result = [];
 		while ( $row = $res->getNext() ) {
 			$result[ $this->uniqid() ] = new ResultItem( $row, $this );
+			usleep( 1 );
```
mwjames: I'm a bit wary about such arbitrary stop process as depends on an environment and may cause non-deterministic behaviour.

s7eph4n: Don't like it either. The problem is, no matter what I try I cannot find any other way to fix this. On a result array of 50 results I get sometimes 8, sometimes 20-ish or 30-ish items, depending on the loop design.

This is on Win7 with PHP 7.1.7 (cli) (built: Jul  6 2017 17:04:27) ( ZTS MSVC14 (Visual C++ 2015) x86 ). If you have any good idea, I'll gladly take it.

mwjames: > On a result array of 50 results I get sometimes 8, sometimes 20-ish or 30-ish items, depending on the loop design.

You mean in `SMWQueryResult` the `next` is returning different results for the same query? Have you tried to verify that `mResults` contains the same result before `next` is applied?

```
	public function getNext() {
		$page = current( $this->mResults );
		next( $this->mResults );
		...
```

`mResults` itself is a simple array (not an iterator nor a generator) so that should not cause any pointer issues.

s7eph4n: Yes, it is always the same query. `mResults` contains 50 entries, but `next` returns `false` too early. Tried resetting `mResults` (which as expected did not help, as it  is the first few entries that are returned. It does not *start* somewhere in between, but *stop* too early), tried replacing `current` + `next` by `each`, even tried not using `getNext` at all, but copy/pasting its content into my loop directly. No dice.

mwjames: > but next returns false too early. 
> but copy/pasting its content into my loop directly. No dice.

Okay, this sounds more of a 7.1 issue, since `mResults` is not a lazy object (something that fetches objects at event point) but a "static" list of entries, I cannot see any other vector that would cause such behaviour.

Issue #292, opened 2017-08-26
Timeline format reports a "Timeline.DefaultEventSource is not a constructor"


mwjames:
### Setup

- MW version:  1.30-alpha
- DB (MySQL etc.): *
- PHP version: *
- SMW version: 2.5.6
- SRF version: 2.5.2 (any version)
- Browsers and versions (if applicable): Chrome

### Issue

![image](https://user-images.githubusercontent.com/1245473/29664220-b55f78e2-8909-11e7-99b4-a02380dcdbad.png)

![image](https://user-images.githubusercontent.com/1245473/29664161-7efb61f8-8909-11e7-80cf-d4ffbf916c00.png)

Most likely relates to:

*  https://github.com/simile-widgets/exhibit/issues/143
* https://github.com/SemanticMediaWiki/SemanticResultFormats/commit/9a2d405d28e18ef50cdda12996d3b260c29e2127
* Async, queue approach of the ResourceLoader

kghbln:
@mwjames Cool! Probably worth backporting?!

kghbln:
The fix was backported to the 2.5.x branch so it will ship with SRF 2.5.3.

PR #293, opened 2017-08-26
Timeline, fix resource loading

mwjames:
This PR is made in reference to: #292, https://sourceforge.net/p/semediawiki/mailman/message/36008850/

- https://github.com/SemanticMediaWiki/SemanticResultFormats/commit/9a2d405d28e18ef50cdda12996d3b260c29e2127 tried to tackle the loading issue in the past but it is believed that recent changes to the ResourceLoader requires a different approach
- Load each file explicitly as outlined in `timeline-api.js`
- Create a `Timeline` instance only after resources have been loaded to avoid a "Timeline.DefaultEventSource is not a constructor"
- Removes the misplaced  `srf-timeline-allresults` with the further results link as is shown even though the JS isn't loaded given a false impression on the current result display or where no further results are available.
- Adds an integration test (doesn't interact with JS, only ensures that data are exported as anticipated by the JS module)

This PR includes:
- [x] CI build passed

Fixes: #292

### DIFF
--- a/Resources.php
+++ b/Resources.php
@@ -516,9 +516,47 @@ return [
 	],
 
 	// Timeline
-	'ext.smile.timeline' => $formatModule + [
-		'scripts' => 'timeline/resources/SimileTimeline/timeline-api.js'
+	// Copied from timeline-api.js
+	'ext.smile.timeline.core' => $formatModule + [
+		'scripts' => [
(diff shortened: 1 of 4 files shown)


kghbln:
Back-ported to 2.5.x with f08b259d6fbfa2387586b1070bd1ad0b67b25e44

Issue #306, opened 2017-09-17
Implement vtimezone in icalendar format


hidrarga:
Hello !

I noticed that the icalendar format does not implement the vtimezone component yet. This is a problem as I use this format to export events on a google calendar. However, the events doesn't appear correctly (for instance with a +2 hours difference). 

I'm willing to implement this, but I would need to know how to obtain timezone information in mediawiki. I tried with $wgLocalTimezone but it always returns null.

Thanks,
HgO

mwjames:
> I'm willing to implement this, but I would need to know how to obtain timezone information in mediawiki. I tried with $wgLocalTimezone but it always returns null

I'm not entirely clear about the use case (TZ of the user setting, TZ of the annotated time, time correction etc.) but in SMW we do have something like [0] and for a stored TZ of an annotated value, the TZ in its serialized form is available via [1] which contains a code and will be resolved during the output by [2].

[0] https://github.com/SemanticMediaWiki/SemanticMediaWiki/blob/master/src/MediaWiki/LocalTime.php
[1]https://github.com/SemanticMediaWiki/SemanticMediaWiki/blob/master/includes/dataitems/SMW_DI_Time.php#L175-L181
[2] https://github.com/SemanticMediaWiki/SemanticMediaWiki/blob/master/src/DataValues/Time/Timezone.php

hidrarga:
>I tried with $wgLocalTimezone but it always returns null.

So, I managed to get this work ! I could implement a generator for the vtimezone component, I will propose a pull request in a few minutes.

>I'm not entirely clear about the use case

My bad, I should have been more precise. For the moment, I'm only assuming that all the events are on the same timezone, that is the server's timezone. Although, in the future, we could change that in order to handle several timezones... Anyhow, thanks for your help ! :)

PR #308, opened 2017-09-29
Adding support for vtimezone in icalendar format

hidrarga:
As stated in #306, the icalendar is missing a vtimezone component. This is an issue with certain calendar applications (such as Google Calendar...), because they will assume the timezone to be UTC when we import the icalendar. 

I propose a small generator of timezone transitions. Because there can be a lot of transitions for a single timezone, I only generate the transitions for the date range of the exported events. See [0] for the vtimezone specifications.

[0] http://www.kanzaki.com/docs/ical/vtimezone.html

This PR includes:
- [ ] CI build passed


mwjames:
Okay, I added #310 before this here goes full throttle which also means that any new functionality should add tests to ensure they are somewhat covered.

mwjames:
>  This is an issue with certain calendar applications (such as Google Calendar...), because they will assume the timezone to be UTC when we import the icalendar.

Can you add a parameter (maybe make it a list parameter to add other transitions in future) so users can choose the make those transitions instead of just assuming that everyone wants that as default?

I haven't really looked at the time code but once you isolate it to a separate class, adding a unit test should be simple matter and makes it easier to test input and output while leaving the icalendar specific formatting separate.

hidrarga:
So, if I well-understood, you would like to have a separate class for the timezone transitions generator ? 

>Can you add a parameter (maybe make it a list parameter to add other transitions in future) so users can choose the make those transitions instead of just assuming that everyone wants that as default?

Did you mean other timezones ? No problem, I'll do that.

I'll do the changes this week when I have some free-time.

mwjames:
> So, if I well-understood, you would like to have a separate class for the
> timezone transitions generator ?

Yes, to ensure we are not doing the kitchen sink approach.

mwjames:
>  commented 19 days ago 

@hidrarga Do we still pursue this PR?

hidrarga:
Sorry, I'm quite overloaded at the moment, but I'd like to have this feature (I applied that patch on my prod wiki, and this was indeed fixing the issue I had with Google Calendar). 

So, I'm working on creating a `SRFTimezones` class (I'm not good with names, so if you have a better one, I'm open to suggestions). However, I can't figure out how to load this new class. I tried with:
```php
$wgAutoloadClasses['SRFTimezones'] = __DIR__ . '/SRF_Timezones.php';
```
But this crashes the wiki. I don't see any error in the apache log, though... Any idea there ?

hidrarga:
@mwjames I pushed my changes so you can maybe tell me what I'm doing wrong. Also, do not hesitate to tell me if this is what you wanted or not :)

mwjames:
Let me add some quick notes:

> Sorry, I'm quite overloaded at the moment, but I'd like to have this feature

This is not a problem.

The issue is that the longer a PR remains unchanged chances are that
is not going to be revived (our experience shows that) and it would be
a shame on the already added effort.

> figure out how to load this new class. I tried with:
> ```php
> $wgAutoloadClasses['SRFTimezones'] = __DIR__ . '/SRF_Timezones.php';
> ```
> But this crashes the wiki. I don't see any error in the apache log,
> though... Any idea there ?

We don't register classes and instead use the composer autoloader to
do the work [0]. In case the class isn't readily available doing
`composer dump-autoload` from the MediaWiki root directory should fix
that (currently we rely on a classmap for those old classes).

Some minor comments on the coding practice:

- No need to add `m_` to variables for new classes
- Rename `m_tzid` to `localTimezone` for a clearer naming
- Add a `setLocalTimezone` to override `localTimezone` during testing
- `from`, `to` are variable actors which the class acts on therefore
instead of invoking them in the constructor, use them as an argument
in something like calcTimezone( from, to ) and later retrieve the
values using `getTimezone`
- Define the `SRFTimezones` instance at the top of `...::getIcal` so
that the `$this->timezone` variable is always instantiate when making
a reference to

In case you have no practice in writing a unit test, at least describe
the expected behaviour of new `SRFTimezones` class such as:

- SRFTimezones::setLocalTimezone( ? ) is set to ... (here we avoid
hacking "wgLocalTimezone")
- SRFTimezones::calcTimezone( ?from, ?to ) with ?from being ..., ?to being ...
- SRFTimezones::getTimezone() should output ...

> 1) SRF\Tests\Integration\JSONScript\JsonTestCaseScriptRunnerTest::executeTestCases with data set "icalendar-02.json" ('/home/travis/build/SemanticMe...2.json')
Failed "#0 `format=icalendar`, match iCalendar specific labels" for
StringContains:

As to why the integration test failed, it is because new fields like
`VTIMEZONE ...` are added to the format and violates the form it is
compared to [1, 2].

```
BEGIN:VTIMEZONE
TZID:UTC
BEGIN:STANDARD
DTSTART:-2922770226570127T082952
TZOFFSETFROM:+0000
TZOFFSETTO:+0000
TZNAME:UTC
END:STANDARD
END:VTIMEZONE
```

> Can you add a parameter (maybe make it a list parameter to add other transitions in future) so users can choose ...

If a timezone parameter is added then the existing integration tests
should not fail for cases where a timezone information is not
generated by default and leaving the format as-is for queries that do
not make use of such parameter.

[0] https://github.com/SemanticMediaWiki/SemanticResultFormats/blob/master/composer.json#L59-L71.

[1] https://github.com/SemanticMediaWiki/SemanticResultFormats/blob/master/tests/phpunit/Integration/JSONScript/Fixtures/icalendar-02.0.txt

[2] https://travis-ci.org/SemanticMediaWiki/SemanticResultFormats/jobs/291291750

Cheers

hidrarga:
Thanks for your help ! 

Indeed, I had to use composer in order to make the autoload work.

I followed your comments, and I also added the possibility to generate transitions for multiple timezones. We can use the `$wgSRFTimezones` global variable to specify from which timezones we want to generate transitions. As you requested, when this variable is not set or is false, the vtimezone components are just not generated. I'd like this variable to default on the same value as `$wgLocalTimezone` (well, this would need to be converted into an array of one element). How could I do that ? 

For the integration tests, I'm not used to phpunit actually. That would be nice if we could enable this feature by default. But then we would need to update the integration tests you mentioned.

Let me know if you need other changes or some clarifications on what I've done :)

hidrarga:
>I'd like this variable to default on the same value as $wgLocalTimezone (well, this would need to be converted into an array of one element). How could I do that ?

Nevermind, I figured out how to do that :) By default it still set to false, but that would be nice if it was set to true.

mwjames:
> Let me know if you need other changes or some clarifications on what I've done :)

There are some style issues (missing parenthesis on if ... else ) and the introduction of the a global wasn't what I had in mind, more precisely, I thought we use:

```
		$params['timezone'] = [
			'default' => '',
			'message' => 'srf-paramdesc-icalendar-timezone',
		];
```

Instead of me trying to use this as an educational exercise, I will merge this PR and make a follow-up. Please have look at those PR's.

> For the integration tests, I'm not used to phpunit actually. That would be nice if we could enable this feature by default. But then we would need to update the integration tests you mentioned.

Please see the follow-ups.

mwjames:
> For the integration tests, I'm not used to phpunit actually. That would be nice if we could enable this feature by default. 

For details, see #338 on how this is done.

PR #331, opened 2017-10-22
Filtered: Use less instead of css

s7eph4n:
Filtered: Use less instead of css

This PR includes:
- [ ] CI build passed


gesinn-it:
@s7eph4n could it be, that this changed the order, when the styles are loaded? We encounter issues, that custom less styles defined via Chameleon now get's loaded before the SRF styles, overrriding our custom styles.

gesinn-it:
Ah, indeed, must have overlooked the nesting. We had

```
.panel {
  .filtered-filters {
```

After changing to 
```
.panel {
  .filtered {
    .filtered-filters {
```

everything works as expected. Thx @s7eph4n .

PR #332, opened 2017-10-22
Load exhibit map assets over HTTPS

TK-999:
- Some assets in Exhibit are always loaded over HTTP. This can cause mixedcontent issues if the wiki is using HTTPS. This change ensures the assets are always loaded over HTTPS.

This PR includes:
- [x] CI build passed

### DIFF
--- a/formats/Exhibit/exhibit/extensions/map/map-extension-bundle.css
+++ b/formats/Exhibit/exhibit/extensions/map/map-extension-bundle.css
@@ -6,5 +6,5 @@ div.exhibit-mapView-map {
     border: 1px solid #738699;
 }
 .olPopupCloseBox {
-    background: transparent url(http://api.simile-widgets.org/ajax/2.2.1/images/close-button.png) no-repeat scroll 0 0 ! important;
+    background: transparent url(https://api.simile-widgets.org/ajax/2.2.1/images/close-button.png) no-repeat scroll 0 0 ! important;
 }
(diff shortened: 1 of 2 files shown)


s7eph4n:
I like it. I think it would be even better to replace these by base64 data-URIs. That would eliminate the requests to an external server completely.
What do you think?

s7eph4n:
See
* Rationale: https://css-tricks.com/data-uris/
* Encoder: https://websemantics.uk/tools/image-to-data-uri-converter/

TK-999:
@s7eph4n yeah, I think that makes perfect sense 🙂 The version in the URL is already hardcoded, so we have only to gain from such a change.

mwjames:
> I think that makes perfect sense :slightly_smiling_face: The version in the URL is already hardcoded

Do we want to change this, or should just I just merged it?

kghbln:
Ping @TK-999

JeroenDeDauw:
We can already merge this as it is a step in the right direction that does not make further improvements as suggested more difficult. So by all means submit a follow up PR.

TK-999:
Thanks :) Sorry about the delay, I was out of office and thus unable to follow up on the status of this PR. I'll submit a subsequent one to convert the assets to base64 encoding and remove dependency on an external web service.

kghbln:
No worries. Thanks anyways for detecting this issue!

kghbln:
Back ported with a2f92dd2fbb0f438779bcdd457a34885c9fae949

Issue #355, opened 2017-11-23
Missing namespace in tagcloud format


cicalese:
### Setup

MediaWiki | 1.29.2 (2b3d6d1)23:58, 21 November 2017
PHP | 5.6.30-0+deb8u1 (apache2handler)
MariaDB | 10.0.33-MariaDB-1~jessie
Semantic MediaWiki | 3.0.0-alpha (6f9c58c)08:25, 20 November 2017 (and earlier)
Semantic Result Formats | 3.0.0-alpha (8fc0c33)08:25, 20 November 2017 (and earlier)

### Issue

When using the tagcloud format with a template to display results that are namespaced pages, the namespace is missing from the results. 

**Steps to reproduce**
This is demonstrated at https://sandbox.semantic-mediawiki.org/wiki/Tag_Cloud_Test.

**Expected result:** 
The results should be the same with and without the template.

**Observed result:** 
The templated results are missing the namespace.

kghbln:
Thanks a lot for reporting this issue and even providing a pull request to fix. I will request a review.

kghbln:
Back-ported to 2.5.x with 34b43fd6a6a92feb8e6490efa55b9e9ee21aa566

PR #356, opened 2017-11-23
Fix missing namespace in result

cicalese:
This PR is made in reference to: #355


mwjames:
Thanks for the fix and to avoid this being dragged I'll merged this without any additional test requirement.
 
PS: We should still aim at adding related tests which in case I will do to avoid going back and forth but I would ask that if time permits you review the test case.

cicalese:
Thank you, @mwjames. I'd be happy to review any test cases.

kghbln:
Back-ported to 2.5.x with 34b43fd6a6a92feb8e6490efa55b9e9ee21aa566

PR #358, opened 2017-11-30
Maintenance on messages

Seb35:
This PR is made in reference to: -

- srf-error-jqplot-bubble-data-length was missing (added in bfe3e83a but not defined)
- smw_qui_limt is unused in SRF and removed from SMW since 6f7625f

This PR includes:
- [ ] CI build passed


Seb35:
If you want I open a bug for this and/or add a line in RELEASE-NOTES.md, I can do this.

s7eph4n:
Thanks for the PR. We don't need a bug, but we'd need a change not only to en.json, but also to qqq.json. That file contains a description of the message helping the translators at translatewiki to understand its context.

Seb35:
Right, I forgot qqq.json. Added.

s7eph4n:
Thanks!

PR #365, opened 2018-01-02
Update for new GraphViz extension class structure

samwilson:
This PR is made in reference to: https://phabricator.wikimedia.org/T181926

- Adds extra check for whether the GraphViz extension is loaded.
- Parses the graphviz source in a more future-compatible way (i.e. the `<graphviz>` element, rather than directly calling the render method).

This PR includes:
- [x] Update of RELEASE-NOTES.md
- [x] CI build passed

### DIFF
--- a/formats/graphviz/SRF_Graph.php
+++ b/formats/graphviz/SRF_Graph.php
@@ -93,7 +93,10 @@ class SRFGraph extends SMWResultPrinter {
 	}
 	
 	protected function getResultText( SMWQueryResult $res, $outputmode ) {
-		if ( !is_callable( 'GraphViz::graphvizParserHook' ) ) {
+		global $wgParser;
+		if ( !class_exists( 'GraphViz' )
+			&& !class_exists( '\\MediaWiki\\Extension\\GraphViz\\GraphViz' )
+		) {
 			wfWarn( 'The SRF Graph printer needs the GraphViz extension to be installed.' );
 			return '';
 		}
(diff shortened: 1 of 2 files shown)


JeroenDeDauw:
Is this compatible with MW 1.27?

samwilson:
Good point! I forgot about the parser service. Switched now; see what you think.

formats/graphviz/SRF_Graph.php
```diff
@@ -112,7 +115,7 @@ protected function getResultText( SMWQueryResult $res, $outputmode ) {
 		$graphInput .= "}";
 		
 		// Calls graphvizParserHook function from MediaWiki GraphViz extension
-		$result = GraphViz::graphvizParserHook( $graphInput, "", $GLOBALS['wgParser'], "" );
+		$result = $wgParser->recursiveTagParse( "<graphviz>$graphInput</graphviz>" );
```
JeroenDeDauw: `$GLOBALS['wgParser']` is better in the sense that you do not pollute the whole function scope with the global

mwjames:
Does this change implies that GraphViz 3.+ is required?

samwilson:
@mwjames: Thanks! And no, it works for old or new versions of GraphViz.

kghbln:
> Does this change implies that GraphViz 3.+ is required?

> Thanks! And no, it works for old or new versions of GraphViz.

@samwilson @mwjames 

FYI [However more admin work.](https://phabricator.wikimedia.org/T181926#3880452)

Issue #373, opened 2018-02-20
Filtered format: select2 list elements rendering outside of containing element


matthew-a-thompson:
### Setup

- MW version: 1.29.0
- DB (MySQL etc.): MySQL 5.6.39
- PHP version: 5.5.38
- SMW version: 2.5.5
- SRF version: 2.5.5-alpha (aad259d)
- Browsers and versions (if applicable): IE11, Chrome

### Issue

When a value filter is used in the filtered result format, it can be displayed either as a list of checkboxes or as a scrollable menu that adds selected elements to a list (a jquery select2 element.) When displaying as a select2 element, the li elements representing each selection are rendering outside of their container.

**Steps to reproduce**

IE11: From the output of a filtered format query with a select2 element, select an item from the select2 list.
Chrome: From the output of a filtered format query with a select2 element, select items from the select2 list. The issue will appear once the total width of selected elements is sufficiently large (greater than the width of the containing element.)

**Expected result:**

li elements with the names of the selected items appear within the containing ul element.

**Observed result:**

li elements appear almost entirely outside the containing ul element, and are only barely visible. See attached screenshots for the observed result in IE11 and Chrome.

In IE11, with the li element barely visible on the left edge of the containing ul element:
![image](https://user-images.githubusercontent.com/22081959/36452918-466dfe66-1664-11e8-92f1-9b42fcfb492f.png)

In IE11, with the li element highlighted:
![image](https://user-images.githubusercontent.com/22081959/36452863-1d71cc68-1664-11e8-8cbd-f31b63ca3079.png)

In Chrome, with the li element barely visible on the left edge of the containing ul element:
![image](https://user-images.githubusercontent.com/22081959/36452936-572bc346-1664-11e8-8005-26364929d1e8.png)

In Chrome, with the li element highlighted:
![image](https://user-images.githubusercontent.com/22081959/36452949-6045bc84-1664-11e8-89e6-54e5dfb31a56.png)

**Proposed solution:**
[...]

PR #374, opened 2018-02-20
Filtered format: select2 list elements rendering outside of containing element

matthew-a-thompson:
Remove selector for .select2-container .select2-search--inline, which was causing select2 elements to render incorrectly.

This PR is made in reference to: #373


kghbln:
This one is missing in master. Not sure if this is meant to be like this.

PR #384, opened 2018-03-21
Updated fullcalendar and added list views

nischayn22:
This PR addresses or contains:
- Updated fullcalendar to the latest version. (FullCalendar v1.5.4 -> v3.9.0)
- Added [list views](https://fullcalendar.io/releases/fullcalendar/3.9.0/demos/list-views.html) to event calendar format

Fixes: #252

PR #385, opened 2018-03-24
Follow-up fullcalendar v3.9.0 API changes, refs 384

mwjames:
This PR is made in reference to: #384

This PR includes:
- [ ] CI build passed


nischayn22:
I have added a new example to https://sandbox.semantic-mediawiki.org/wiki/Issue/SRF/384
The following configuration doesn't work: views=basicday,month,listday,listmonth,listweek

When we are doing a replace of 'day' with 'Day' only the first occurrence is getting replaced in this example basicday becomes basicDay but listday remains listday and doesn't show up.

mwjames:
>The following configuration doesn't work: views=basicday,month,listday,listmonth,listweek
> When we are doing a replace of 'day' with 'Day' only the first occurrence is getting replaced in this example basicday becomes basicDay but listday remains listday and doesn't show up.

I suspect https://github.com/SemanticMediaWiki/SemanticResultFormats/commit/983428f5046230e2af02e75abb0c0a17fb6f8d39 should fix that.

nischayn22:
Tested that it works. Thanks!

PR #395, opened 2018-06-16
Correctly format whole-hour timezones west of UTC

raboof:
The change *seems* OK to me, but I wasn't able to run the unit tests locally and it isn't clear from the Travis output what might be wrong.

I'd love some guidance on this, otherwise feel free to close this PR.

(the next step would be to add a test with the `America/St_Johns` timezone, as I suspect there is an `abs` needed there as well, but I wanted to make one change at a time)


kghbln:
Thanks for the pull! 

@JeroenDeDauw It will be great if you could have a peep at this.

kghbln:
@JeroenDeDauw Re-ping.

kghbln:
Looking at this makes me a bit sad.

PR #409, opened 2018-08-25
Fix "supress" typo in jquery.blockUI.js file

zoranzoki21:


zoranzoki21:
@JeroenDeDauw Can you merge and that? :)

kghbln:
That's another upstream issue. The pull should be directed to https://github.com/malsup/blockui/blob/master/jquery.blockUI.js Once we move in an update the fix will end up here, too.

Thanks again for working on these tiny typos!

PR #435, opened 2018-09-30
IcalTimezoneFormatter fix tz-transitions, refs 392, 395

### LINKED ISSUE #392: iCalendar format: Avoid 'floating' times in export
### Setup

- MediaWiki | 1.30.0 (636a2b9)22:27, 30 April 2018
- PHP | 7.0.27-0+deb9u1 (apache2handler)
- MariaDB | 10.1.26-MariaDB-0+deb9u1
- Semantic MediaWiki | 3.0.0-alpha (497c969) 20:17, 29 April 2018
- Semantic Result Formats | 3.0.0-alpha (540e63e) 09:28, 19 April 2018

### Issue

The iCalendar export now includes a timezone section, but start/end times are still represented as 'floating' times. Those have really weird semantics in iCalendar (https://tools.ietf.org/html/rfc5545#section-3.3.12), so it would be good to avoid that.

https://www.semantic-mediawiki.org/wiki/Help:Type_Date notes SMW times are in the server/wiki timezone. Suppose the server/wiki timezone is GMT+1, an events that starts at "2019-01-12 18:00" is now represented in iCalendar format as `DTSTART:20190112T083000`.

If we'd wanted to avoid 'floating' times, we have 2 options: we could convert the date to UTC (leading to `DTSTART:20190112T073000Z`) or add a `TZID` parameters (leading to `DTSTART;TZID=Europe/Paris:20190112T083000`).

I just tested and both variations work as expected when importing into Google Calendar (unlike the current 'floating' time).

I'd be happy to contribute this feature. I'm on the fence whether I'd prefer to implement exporting as UTC or exporting with a TZID parameter, it's hard to say which is more likely to be interpreted 'correctly' even by importers that do not implement RFC5545 completely.

Any preference?

mwjames:
This PR is made in reference to: #392, #395

- Use the approach from #395 but restrict available transitions so that it works in both directions

This PR includes:
- [x] CI build passed

Fixes #392, Fixes #395


kghbln:
Thanks a ton!

PR #443, opened 2018-10-12
Tree format

### LINKED ISSUE #287: Tree format <div> container breaks TreeAndMenu?
### Setup

- MW version: 1.29.0 (on a private domain)
- DB (MySQL etc.): 10.1.21-MariaDB
- PHP version: 5.6.30
- SMW version: 2.5.4
- SRF version: 2.5.1
- Browsers and versions (if applicable):

### Issue

I use the SRF tree format in combination with the [TreeAndMenu](https://www.mediawiki.org/wiki/Extension:TreeAndMenu) extension to build an interactive overview of a thesaurus. After upgrading MW and SMW, this no longer works.

**Steps to reproduce**
1. Install TreeAndMenu
2. Create an ultree (as in https://sandbox.semantic-mediawiki.org/wiki/Ulree_example) wrapped in a TreeAndMenu #tree. The example from my site is
`{{#tree:root=Thesaurus|
{{#ask:
[[Category:Thesaurus]]
[[Is narrower than::+]]
 | format=tree
 | limit=200
 | parent=Is narrower than
}}
}}`

**Expected result:** ...
The whole tree is rendered.
**Observed result:** ...
Only the Thesaurus root node (specified after #tree) is shown.

**My own solution**
The problem seems to be that the SRF tree for 2.5.1 is wrapped in a HTML div container, with class srf-tree (formats/tree/TreeResultPrinter.php, line 149), while TreeAndMenu expects a ul list. My previous version of SRF didn't put it in the div container AFAIK. I solved this issue for now by removing the Html:rawElement method and simply returning the result.
`return Html::rawElement( 'div', [ 'class' => 'srf-tree' ], $resultText );`
becomes
`return $resultText;`

I'm not sure what the rationale behind the div container was, and whether this is a SRF or TreeAndMenu bug. If it is the latter please let me know and I will log the bug there.

s7eph4n:
This PR is made in reference to: #287

If the "class" parameter is present, the result will be wrapped inn a div with the specified class.  Otherwise the result will be a plain ul or ol.

This PR includes:
- [x] CI build passed

Fixes #287


kghbln:
Also to be back-ported to 3.0.x branch.

formats/tree/TreeResultPrinter.php
```diff
@@ -137,7 +137,7 @@ protected function getResultText( SMWQueryResult $queryResult, $outputmode ) {
 
 		$this->setQueryResult( null );
 
-		return Html::rawElement( 'div', [ 'class' => 'srf-tree' ], $resultText );
+		return $this->params[ 'class' ] ? Html::rawElement( 'div', [ 'class' => $this->params[ 'class' ] ], $resultText ) : $resultText;
```
JeroenDeDauw: Better to do a strict check and you can use a guard clause to make things more readable:

```suggestion
		if ( trim( $this->params[ 'class' ] ) === '' ) {
			return $resultText;
		}
		
		return Html::rawElement( 'div', [ 'class' => $this->params[ 'class' ] ], $resultText );
```

s7eph4n:
Thanks!

kghbln:
> Also to be back-ported to 3.0.x branch.

That's actually an enhancement, thus SRF 3.1.0

Issue #396, opened 2018-07-05
Gallery format: query parameter `default` is being ignored


kghbln:
### Setup
- MediaWiki | 1.31.0 (32d551a)01:34, 2. Jul. 2018
- PHP | 7.0.27-0+deb9u1 (apache2handler)
- MariaDB | 10.1.26-MariaDB-0+deb9u1
- Semantic MediaWiki | 3.0.0-alpha (775811b) 01:44, 4. Jul. 2018
- Semantic Result Formats | 3.0.0-alpha (1093a59) 08:06, 7. Mai 2018

### Issue
In case an inline query does not yield a result the text provided via the `default` query parameter is not shown.

See the [example](https://sandbox.semantic-mediawiki.org/wiki/Gallery) on sandbox.

kghbln:
@JeroenDeDauw FYI Since you did substantial work on this result format ...

PR #455, opened 2018-12-15
Gallery format: fix default= setting

krabina:
This PR is made in reference to: #396 

new private function getNumbers
call to function right after getResultText
function addNumbersForDataItem (was missing pn previous PR)

Sorry for my unprofessional pull request :-)

This PR includes:
- [ ] CI build passed

Fixes #

### DIFF
--- a/formats/gallery/Gallery.php
+++ b/formats/gallery/Gallery.php
@@ -65,7 +65,7 @@ class Gallery extends ResultPrinter {
 	 */
 	public function getResultText( SMWQueryResult $results, $outputmode ) {
 		$numbers = $this->getNumbers( $results );
-				if ( count( $numbers ) == 0 ) {
+				if ( $results->getCount() === 0 ) {
 					return $this->params['default'];
 				}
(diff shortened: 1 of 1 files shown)


krabina:
i copied the code from math result, by the way

s7eph4n:
I don't understand the logic of this PR. How is it supposed to work?
And there is a relevant test failing (see travis), so either this patch does not do what it should, or the test needs fixing.

krabina:
sorry about that. as I said I copied code from the math result format. it works in my wiki, but I did not test it professionally. thought it must be an easy fix as it works in the other result formats. Feel free to just delete it.

kghbln:
@krabina I guess this pull can be closed?

krabina:
I guess so

kghbln:
> I guess so

Happened to me too. Trying to imitate something existing because it seemingly looks identical in a way and end up in a ball of issues because things are not similar at all. To cut it short: I feel with you.

PR #457, opened 2019-01-12
Add `template` to outline format, refs 88

### LINKED ISSUE #88: Outline format: add support for a "template=" parameter
Moved here from: https://github.com/SemanticMediaWiki/SemanticMediaWiki/issues/849
- 'outline' format - add support for a "template=" parameter, similar to the ones found in various other formats. [independent]

mwjames:
This PR is made in reference to: #88 

- Adds the `template` parameter and `TemplateBuilder` to create individual hierarchical output lists using a template
- `named args` are `true` by default and I haven't considered implementing numerical parameter (aka `{{{1}}}`) support in the `TemplateBuilder` (which means it is missing by design and if required someone should send a patch and test)
- Adds `outline-01.json` integration test to verify that the `template` produces a specified output
- The format hasn't seen any updates in years and this PR fixes a notorious "Notice: Array to string conversion in .../w/extensions/SemanticMediaWiki/includes/SMW_Infolink.php on line 469" caused by the wrong use of `$link = $res->getQueryLink(); / $link->setParameter( $this->params['outlineproperties'], 'outlineproperties' );`

This PR includes:

- [x] CI build passed

Fixes #88 

## Example

Using the `template` output allows to create lists like the phabricator task list (To create that output was the reason I took an interest in this format in the first place.).

image

### DIFF
--- a/formats/outline/SRF_Outline.php
+++ b/formats/outline/SRF_Outline.php
@@ -1,5 +1,7 @@
 <?php
 
+use SRF\Outline\TemplateBuilder;
+
 /**
  * A class to print query results in an outline format, along with some
  * helper classes to handle the aggregation
@@ -43,6 +45,8 @@ class SRFOutlineTree {
 
 	var $mTree;
 	var $mUnsortedItems;
+	var $itemCount = 0;
+	var $leafCount = 0;
 
 	function __construct( $items = [] ) {
 		$this->mTree = [];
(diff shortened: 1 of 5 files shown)


mwjames:
I added [0] which includes the template and once the PR is available should produce above example output. I think this is mostly relevant for users who mange task or project lists (FYI @kghbln, @krabina, @s7eph4n).

[0] https://sandbox.semantic-mediawiki.org/wiki/SRF/Outline/Template/PhabListView

mwjames:
@kghbln Maybe give the sandbox a nudge to fetch the SRF master.

kghbln:
> Maybe give the sandbox a nudge to fetch the SRF master.

Will do in a quarter of an hour. Currently it still appears to need some more time to infiltrate.

kghbln:
> Will do in a quarter of an hour. Currently it still appears to need some more time to infiltrate.

I am afraid this takes much longer or has an issue as encountered yesterday for SMW.

PR #477, opened 2019-03-20
Use SMW copy of jquery.async

Krinkle:
Ref https://github.com/SemanticMediaWiki/SemanticMediaWiki/pull/3823.


kghbln:
Thanks for the pull!

PR #496, opened 2019-07-15
Add `nodelabel` parameter with `displaytitle` option

jewloszewicz:
- Adds `nodelabel` parameter with `displaytitle` option based on [DISPLAYTITLE](https://www.mediawiki.org/wiki/Extension:Display_Title). It allows to display correct page titles on graph nodes.

This PR includes:
- [ ] CI build passed

### DIFF
--- a/formats/graphviz/SRF_Graph.php
+++ b/formats/graphviz/SRF_Graph.php
@@ -15,6 +15,12 @@
  */
 class SRFGraph extends SMWResultPrinter {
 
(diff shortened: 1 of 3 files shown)


mwjames:
The failing tests are not caused by this PR, also maybe use something
like `'values' => [ 'displaytitle' ]` as parameter value instead of a
free text since it is an option/list field.

```
		$params['nodelabel'] = [
			'default' => [],
			'values' => [ 'displaytitle' ],
		];
```

mwjames:
Thanks

PR #297, opened 2017-08-31
Graph format: Refactoring and enhancing

gesinn-it:
This PR is made in reference to: #

This PR addresses or contains:
- refactored, added SRFGraphNode class
- added handling for additional labels (label1 - label3)
- added record/Mrecord type shapes
- added arrowhead parameter
- added support for Display Title

This PR includes:
- [ ] Update of RELEASE-NOTES.md
- [ ] Tests (unit/integration)
- [ ] CI build passed

Issue #63, opened 2014-07-21
Next release


JeroenDeDauw:
Let's go with `2.0` together with SMW, so SRF is also semver compliant?

PR #507, opened 2019-07-26
Refactoring, Add UnitTest, Add graphlabel  (commit-ID Github: bb7ff5e…

RacoonDev:
…9a3d224516d9dc0295f6e640e638a551f)

This PR is made in reference to: 
- #63,
- #297 

- Refactoring
- Adjust Unit Tests
- Use of nodelabel

This PR includes:
- [ ] CI build passed


src/Graph/GraphNode.php
```diff
@@ -0,0 +1,50 @@
...
+
+class GraphNode {
+	private $id;
+	private $parent = [];
+	private $label;
+
+	/**
+	 * @var string $id : Node ID including namespace
+	 */
+	public function __construct( $id ) {
```
JeroenDeDauw: Meh. What about bumping the min PHP version of SRF to 7.1? Travis already at 7.0 so not much of a change. @mwjames @kghbln @RacoonDev

src/Graph/GraphPrinter.php
```diff
@@ -101,183 +112,203 @@ class GraphPrinter extends SMWResultPrinter {
...
 			&& !class_exists( '\\MediaWiki\\Extension\\GraphViz\\GraphViz' )
 		) {
 			wfWarn( 'The SRF Graph printer needs the GraphViz extension to be installed.' );
 			return '';
 		}
 
-		$this->isHTML = true;
+		///////////////////////////////////
+		// GRAPH OPTIONS
+		///////////////////////////////////
```
JeroenDeDauw: Not like this. If the code is not clear try extract well named functions.

src/Graph/GraphPrinter.php
```diff
@@ -101,183 +112,203 @@ class GraphPrinter extends SMWResultPrinter {
...
+
+		$graphInput = "digraph $this->graphName {";
+
+		// fontsize and fontname
+		$graphInput .= "graph [fontsize=10, fontname=\"Verdana\"]\nnode [fontsize=10, fontname=\"Verdana\"];\nedge [fontsize=10, fontname=\"Verdana\"];";
 
-		$graphInput = "digraph $this->m_graphName {";
-		if ( $this->m_graphSize != '' ) {
-			$graphInput .= "size=\"$this->m_graphSize\";";
+		// size
```
JeroenDeDauw: ???

src/Graph/GraphPrinter.php
```diff
@@ -101,183 +112,203 @@ class GraphPrinter extends SMWResultPrinter {
...
+	/**
+	 * Process a result row and create SRF\GraphNodes
+	 *
+	 * @since 2.5.0
+	 *
+	 * @param array $row
+	 * @param $outputmode
+	 * @param array $nodes
+	 *
+	 */
```
JeroenDeDauw: Comment is wrong

src/Graph/GraphPrinter.php
```diff
@@ -343,32 +367,32 @@ public function getParamDefinitions( array $definitions ) {
 		];
 
 		$params['graphsize'] = [
-			'type' => 'string',
-			'default' => '',
-			'message' => 'srf-paramdesc-graphsize',
+			'type'              => 'string',
+			'default'           => '',
+			'message'           => 'srf-paramdesc-graphsize',
```
JeroenDeDauw: Don't vertical align please

tests/phpunit/Unit/Graph/GraphTest.php
```diff
@@ -0,0 +1,78 @@
...
+
+use SMW\Test\QueryPrinterRegistryTestCase;
+use SRF\Graph\GraphNode;
+
+
+/**
+ * Tests for the SRF\Graph class.
+ *
+ * @file
+ * @since 1.8
```
JeroenDeDauw: Wrong

tests/phpunit/Unit/Graph/GraphTest.php
```diff
@@ -0,0 +1,78 @@
...
+ *
+ * @ingroup SemanticResultFormats
+ * @ingroup Test
+ *
+ * @group SRF
+ * @group SMWExtension
+ * @group ResultPrinters
+ *
+ * @licence GNU GPL v2+
+ * @author Jeroen De Dauw < jeroendedauw@gmail.com >
```
JeroenDeDauw: Wrong

tests/phpunit/Unit/Graph/GraphTest.php
```diff
@@ -0,0 +1,78 @@
+<?php
+
+namespace SRF\Tests\Unit\Formats;
+
+use SMW\Test\QueryPrinterRegistryTestCase;
+use SRF\Graph\GraphNode;
+
+
+/**
+ * Tests for the SRF\Graph class.
```
JeroenDeDauw: Wrong

Instead: `@covers SRF\Graph\GraphPrinter`

tests/phpunit/Unit/Graph/GraphTest.php
```diff
@@ -0,0 +1,78 @@
...
+
+	/**
+	 * @see QueryPrinterRegistryTestCase::getClass
+	 *
+	 * @since 1.8
+	 *
+	 * @return string
+	 */
+	public function getClass() {
+		return 'SRF\Graph\GraphPrinter';
```
JeroenDeDauw: `GraphPrinter::class` (don't forget to import it)

tests/phpunit/Unit/Graph/GraphTest.php
```diff
@@ -0,0 +1,78 @@
...
+	}
+
+	/**
+	 * Testing class GraphNode
+	 *
+	 * @since 3.0
+	 *
+	 */
```
JeroenDeDauw: Wrong again. This comment is not needed anyway like so many of the others. Not adding them is much better.

tests/phpunit/Unit/Graph/GraphTest.php
JeroenDeDauw: Hold on... You are actually testing the GraphNode class... Then why are you extending `QueryPrinterRegistryTestCase` and implementing a bunch of stuff you don't need?

mwjames:
> Meh. What about bumping the min PHP version of SRF to 7.1? Travis already at
> 7.0 so not much of a change. @mwjames @kghbln @RacoonDev

If we want to keep it easy for users (not for developers or CIs) then
we should aim at the same boundaries as defined by SMW 3.1 [0].

[0] https://github.com/SemanticMediaWiki/SemanticMediaWiki/issues/3918#issuecomment-482940823

tests/phpunit/Unit/Graph/GraphTest.php
```diff
@@ -0,0 +1,78 @@
...
+
+		$mockParentNode1[] = [
+			"predicate" => 'Part Of Team',
+			"object"    => 'Alpha Team'
+		];
+
+		$node->addParentNode( 'Part Of Team', 'Alpha Team' );
+		$this->assertEquals( $mockParentNode1, $node->getParentNode() );
+
+	}
```
JeroenDeDauw: Please follow http://wiki.c2.com/?ArrangeActAssert. You can create more than one test method.

mwjames:
>
> Wrong again. This comment is not needed anyway like so many of the others.
> Not adding them is much better.

Here we disagree, I'd like see at which point (release) a public
method was added. This is especially helpful to users unfamiliar wit
the code base, and no, git blame is not substitution for some minimal
code documentation (the same goes for strict typing).

JeroenDeDauw:
@mwjames I doubt we disagree. The method I commented on is in a test (hence package private) which I figure you did not see in the email you replied to.

JeroenDeDauw:
@RacoonDev I was tired yesterday when making this review so its sloppy and lazy. Thanks for the changes. I do appreciate some of the refactoring, however there are a bunch of code quality issues to be solved.

JeroenDeDauw:
> If we want to keep it easy for users (not for developers or CIs) then
> we should aim at the same boundaries as defined by SMW 3.1 [0].

@mwjames I think this syncing of version numbers and same dependencies between SMW and SRF made sense back in the day, though it's not clear to me it provides value now installation happens with Composer.

mwjames:
> @mwjames I think this syncing of version numbers and same dependencies
> between SMW and SRF made sense back in the day, though it's not clear to me
> it provides value now installation happens with Composer.

I think Karsten has some insights on the matter of 7.0 vs 7.1
(discussed somewhere on the SMW issue list) with view on the
Debian/Ubuntu release process. It is not so much about Composer rather
then for users to understand the range is simply MW 1.31 - 1.33 PHP
7.0 - 7.3 and with that consistent rather than having for each package
different requirements.

I see no specific reason (meaning from a technical/security
perspective) to require PHP 7.1 just for SRF 3.1 because it is PHP
7.1. Of course, if there are some emergent security or technical
objectives then is is a different matter.

[0] https://www.mediawiki.org/wiki/MediaWiki_1.31 "MediaWiki 1.31
requires PHP 7.0.0 or later. "

kghbln:
We have discussed the version requirements at different spots already. Anyways, since SMW 3.1 is meant to support MW 1.31 we should really use their minimum requirement, too.  Moreover people still running Debian 9 (Debian 10 was released just a fortnight ago.) will not be able to use SMW 3.1.0 and SRF 3.1.0 if it is PHP 7.1. Happens to me with e.g Maps.. There are features out there for a year in the meantime which I cannot use. So far we were more concerned with withholding features by having a wide release process, but the the over way around is nearly as bad I believe.

>  I think this syncing of version numbers and same dependencies between SMW and SRF made sense back in the day, though it's not clear to me it provides value now installation happens with Composer.

Now it is more a marketing thing, mostly because SRF is still the most important extension to SMW and it did not require any big effort to keep this synced. Also it is is bit easier for server administrators since they only have to adapt their "composer.local.json" once for these two.

mwjames:
@RacoonDev Sorry about the release boundary discussions and the noise which is surely the wrong place here therefore I'll hide those discussion points.

gesinn-it:
@JeroenDeDauw @mwjames many thanks for your reviews and feedback. @RacoonDev is on vacation :-) He will follow up on this PR after his return.

src/Graph/GraphPrinter.php
RacoonDev: I removed those heavy comments and if necessary i outsourced code to own function.

See: https://github.com/RacoonDev/SemanticResultFormats/commit/76983e5c31d988a87b94720aee2215e48ae29886

src/Graph/GraphPrinter.php
RacoonDev: What exactly do you mean here? Please give me a hint

src/Graph/GraphPrinter.php
RacoonDev: see: https://github.com/RacoonDev/SemanticResultFormats/commit/2733ba422c72be6b4d3d7c26720fac7b4fd27c3b

tests/phpunit/Unit/Graph/GraphTest.php
RacoonDev: see: https://github.com/RacoonDev/SemanticResultFormats/commit/c1acdf0cf7f136a108ae0380959954703d0bbd35

tests/phpunit/Unit/Graph/GraphTest.php
RacoonDev: see: https://github.com/RacoonDev/SemanticResultFormats/commit/c1acdf0cf7f136a108ae0380959954703d0bbd35

tests/phpunit/Unit/Graph/GraphTest.php
RacoonDev: see: https://github.com/RacoonDev/SemanticResultFormats/commit/c1acdf0cf7f136a108ae0380959954703d0bbd35

tests/phpunit/Unit/Graph/GraphTest.php
RacoonDev: see: https://github.com/RacoonDev/SemanticResultFormats/commit/c1acdf0cf7f136a108ae0380959954703d0bbd35

tests/phpunit/Unit/Graph/GraphTest.php
RacoonDev: see: https://github.com/RacoonDev/SemanticResultFormats/commit/c1acdf0cf7f136a108ae0380959954703d0bbd35

tests/phpunit/Unit/Graph/GraphTest.php
RacoonDev: see: https://github.com/RacoonDev/SemanticResultFormats/commit/c1acdf0cf7f136a108ae0380959954703d0bbd35

src/Graph/GraphPrinter.php
RacoonDev: see: https://github.com/RacoonDev/SemanticResultFormats/commit/2733ba422c72be6b4d3d7c26720fac7b4fd27c3b

src/Graph/GraphPrinter.php
JeroenDeDauw: I'm not sure why you added this comment. It adds no info, just noise

RacoonDev: Yeah thats right, I'll pay more heed to this in the future. 
see: https://github.com/RacoonDev/SemanticResultFormats/commit/76983e5c31d988a87b94720aee2215e48ae29886

src/Graph/GraphNode.php
RacoonDev: I think i can resolve this one...

> @RacoonDev Sorry about the release boundary discussion and the noise which is surely the wrong place here therefore I'll hide those discussion points.

JeroenDeDauw:
Do you want this new format to be available in a release supporting PHP 7.0 or is 7.1 fine as a requirement?

RacoonDev:
@JeroenDeDauw for me it would be ok if the new format was included with the SRF release for php 7.1.

tests/phpunit/Unit/Graph/GraphNodeTest.php
```diff
@@ -0,0 +1,50 @@
...
+
+class GraphNodeTest extends \PHPUnit_Framework_TestCase {
+
+	public function testCanConstruct() {
+
+		$this->assertInstanceOf(
+			GraphNode::class,
+			new GraphNode( "graphnode:id" )
+		);
+	}
```
JeroenDeDauw: You can drop this test

tests/phpunit/Unit/Graph/GraphNodeTest.php
```diff
@@ -0,0 +1,50 @@
...
+
+	public function testGraphNode(){
+
+		$node = new GraphNode( 'Team:Beta' );
+		$this->assertEquals( 'Team:Beta', $node->getID() );
+
+		$node->setLabel( "Fossil Power Generation" );
+		$this->assertEquals( "Fossil Power Generation", $node->getLabel() );
```
JeroenDeDauw: When comparing scalar values you can use assertSame, which is more strict. (0 and "0" will not be equal)

tests/phpunit/Unit/Graph/GraphNodeTest.php
```diff
@@ -0,0 +1,50 @@
...
+
+use SRF\Graph\GraphNode;
+
+
+/**
+ * @covers \SRF\Graph\GraphNode
+ * @group semantic-result-formats
+ *
+ * @licence GNU GPL v2+
+ * @since 3.1
```
JeroenDeDauw: No need to add @since tags in the tests since they are not package public

src/Graph/GraphPrinter.php
```diff
@@ -87,197 +86,225 @@ class GraphPrinter extends SMWResultPrinter {
...
-				$color = $this->m_graphColors[array_search( $labelName, $this->m_labelArray, true )];
+	/**
+	 * Process a result row and create SRF\GraphNodes
+	 *
+	 * @since 3.1
+	 *
+	 * @param array $row
+	 *
+	 */
+	protected function processResultRow( array /* of SMWResultArray */ $row ) {
```
JeroenDeDauw: This should probably be private. Try using private by default.

Generally speaking very few things are protected. In result printer classes you will need to have a bunch due to the inheritance design though.

If it is private it also should not have a @since tag

src/Graph/GraphPrinter.php
```diff
@@ -87,197 +86,225 @@ class GraphPrinter extends SMWResultPrinter {
...
-					$this->m_labelArray[] = $labelName;
-				}
-
-				$color = $this->m_graphColors[array_search( $labelName, $this->m_labelArray, true )];
+	/**
+	 * Process a result row and create SRF\GraphNodes
+	 *
+	 * @since 3.1
+	 *
+	 * @param array $row
```
JeroenDeDauw: @param SMWResultArray[] $row

src/Graph/GraphPrinter.php
```diff
@@ -87,197 +86,225 @@ class GraphPrinter extends SMWResultPrinter {
...
-	protected function getGVForDataValue( SMWDataValue $object, $outputmode, $isName, $name, $labelName ) {
-		$graphInput = '';
-		$nodeLabel = '';
-		$text = $object->getShortText( $outputmode );
-
-		if ( $this->m_graphLink ) {
-			$nodeLinkURL = "[[" . $text . "]]";
-		}
+	protected function getGraphLegend(){
+		if ( $this->showGraphLegend && $this->showGraphColor ) {
```
JeroenDeDauw: There is no return value outside of this block. Going to lead to errors in strict mode I figure. You can have a guard clause at the start here that returns an empty string.

src/Graph/GraphPrinter.php
```diff
@@ -87,197 +86,225 @@ class GraphPrinter extends SMWResultPrinter {
...
 	}
 
-	protected function getResultText( SMWQueryResult $res, $outputmode ) {
 
+	/**
+	 * @param SMWQueryResult $res
+	 * @param $outputmode
+	 * @return string
+	 */
+	protected function getResultText( SMWQueryResult $res, $outputmode ) {
```
JeroenDeDauw: This function is waaaaay to long

JeroenDeDauw:
@RacoonDev with PHP 7.1 you can add scalar type hints and return types (no need to do this here now if you do not want to... just something I recommend in general)

mwjames:
> That said, I do think perfect is the enemy of the good and this is not
> entirely new code. I'm also not sure what standard we want to use for new
> code in this repo. What is in the PR is not worse than what we have in most
> other formats... @mwjames thoughts on where to set the bar for new code and
> changes?

Yes, refactoring an existing class or format isn't fun but eventually
becomes necessary otherwise no one is able to maintain or improve its
functionality.

As for testing, creating an integration (not unit) test for this
format seems rather difficult as it relies on GraphViz as external
dependency. So, my suggestions would be (after this PR got merged),
you have a look at what happens within the while loop where you create
the formatting rules (with lines like "\"" . $nodeName ...) and
isolate the formatting to a dedicated formatter so that in the end we
can have something like (beware of the pseudo code, don't get fooled
but it should highlight the direction):

```

$options = ...;

while ... {
   ...
   $nodes[] = new GraphNode(... );
}

$graphFormatter = new GraphFormatter(  $options );
$graph = $graphFormatter->buildGraph( $nodes );

$GLOBALS['wgParser']->parse( $graph ...);
```

Above can be tested isolated, from whether GraphViz is installed or
not, and allows you to generate content independently from the
`QueryResult` object. It further separates the GraphViz specific
formatting rules about how the output should look like allowing to be
extendable in future if the need arises.

So, in short, if Travis is green, merged it, and have another look at
the formatter as suggested above as it would surely make the class
more readable.

Cheers

RacoonDev:
Thank you both @mwjames and @JeroenDeDauw for helping me to improve the code quality. I know that the code is kind of nasty but refactoring an existing format is not fun like mwjames said. I try to outsource the functionality of GraphPrinter::getResultText() to its own GraphFormatter-Class as requested so please wait for merging the this PR into the master. 

I came across the "nameproperty" that is mentioned on SRF graph format help site [0]. Somehow the functionality behind this property got lost during the refactoring but to me it looks like that the "nodelabel" [1] that was implemented by @jewloszewicz try to handle the same problem. I'm confused and would be thankful for some advice if i can remove this or am I a victim of my own sloppy code.

[0] https://www.semantic-mediawiki.org/wiki/Help:Graph_format
[1] https://github.com/SemanticMediaWiki/SemanticResultFormats/commit/bb7ff5e9a3d224516d9dc0295f6e640e638a551f

mwjames:
> format is not fun like mwjames said. I try to outsource the functionality of
> GraphPrinter::getResultText() to its own GraphFormatter-Class as requested
> so please wait for merging the this PR into the master.

Let's merge this one first and tackle the formatter issue hereafter
allowing it be an incremental change otherwise you end up with a huge
pile of changes where things like the nodelabel/nameproperty can get
lost.

> I came across the "nameproperty" that is mentioned on SRF graph format help
> site [0]. Somehow the functionality behind this property got lost during the
> refactoring but to me it looks like that the "nodelabel" [1] that was
> implemented by @jewloszewicz try to handle the same problem. I'm confused

Maybe tackle this after the PR is merged (and before the formatter
thing) so that is easily trackable and manually testable (and
verifiable) on what this "nameproperty" should be doing and currently
is not.

Cheers

JeroenDeDauw:
Alright, merging since James is not screaming murder at the suggestion :)

mwjames:
> outsource the functionality of GraphPrinter::getResultText() to its own GraphFormatter-Class as requested

You can find some inspiration in #535 which isolates the export specific formatting for a different format so that it can be tested without having to construct a `QueryResult` object.

Issue #371, opened 2018-02-20
vCard format: introduce `filename` parameter


krabina:
### Setup

- MW version:1.27
- SRF version: 2.5.4

### Issue

In contrast to the csv export format, the vCard format does not know a |filename= parameter. Every result format that delivers a file should have the possibility of specifying a filename.

kghbln:
Thanks for reporting. Use the `searchlabel` paramter for this like e.g.:
```text
Download the vCard of {{#ask:
 [[Category:Person]]
 |?firstname
 |?lastname
 |?mail=email
 |?url=homepage
 |?skype=note
 |format=vcard
 |searchlabel=Bernhard Krabina
}}
```
@krabina Does this work for you?

krabina:
yes, thank you. But it does someting a little different. 
I'd like to have "vCard" displayed and the resulting filename is e. g. Krabina.vcf
Your workaround is fine for me at the moment, but I think this should be streamlined according to the CSV format. (Haven't tried, but I guess, the same thing happens with Excel format).

kghbln:
This is why I did not close this one but instead labeled as enhancement. Same as this one goes for the "bibtex" format.

PR #510, opened 2019-07-27
Tidy `format=bibtex`

mwjames:

- Isolates text formatting and field collection using a separate `Item` class
- Adds the `filename` parameter (https://github.com/SemanticMediaWiki/SemanticResultFormats/issues/371#issuecomment-367017087)

This PR includes:
- [x] CI build passed

### DIFF
--- a/src/BibTex/Item.php
+++ b/src/BibTex/Item.php
@@ -1,0 +1,185 @@
+<?php
+
+namespace SRF\BibTex;
+
+use SMWDataValue as DataValue;
+
+/**
+ * @see http://www.semantic-mediawiki.org/wiki/BibTex
+ *
+ * @license GNU GPL v2+
+ * @since 3.1
+ *
+ * @author mwjames
+ */
+class Item {
+
+	/**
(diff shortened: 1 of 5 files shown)


mwjames:
@kaefer3000 The network overview in Github showed that you made some changes to the bibtex format on a fork, I hereby let you know that those changes have to be replied for the upcoming SRF 3.1 due to a refactored `BibTex` interface.

src/BibTex/Item.php
```diff
@@ -0,0 +1,185 @@
...
+
+		return $text;
+	}
+
+	/**
+	 * Consist of `author last name` + `year` + `first word of title`
+	 *
+	 * @return string
+	 */
+	protected function buildURI() {
```
JeroenDeDauw: Why not private?

mwjames: I had them private first but then I looked at @kaefer3000 (https://github.com/SemanticMediaWiki/SemanticResultFormats/compare/master...aifb-smw:bibtex-3.0.x) and thought we should allow them to override those methods by applying local changes with their own item class.

PR #517, opened 2019-07-29
Fix bug in MapView format

JeroenDeDauw:
This check was still using the old class name which got removed some time ago. Hence this code was broken on modern PHP versions.

Removed the check altogether in second commit since there is no need for it.

### DIFF
--- a/formats/filtered/src/View/MapView.php
+++ b/formats/filtered/src/View/MapView.php
@@ -4,7 +4,6 @@ namespace SRF\Filtered\View;
 
 use DataValues\Geo\Parsers\LatLongParser;
 use Exception;
-use Message;
 use SMWPropertyValue;
 use SRF\Filtered\ResultItem;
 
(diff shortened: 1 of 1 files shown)


formats/filtered/src/View/MapView.php
```diff
@@ -78,11 +77,6 @@ public function getJsDataForRow( ResultItem $row ) {
 						}
 					}
 
-				} else {
-					$this->getQueryPrinter()->addError(
-						Message::newFromKey( 'srf-filtered-map-geocoordinateparser-missing-error' )->inContentLanguage(
```
s7eph4n: I think this is the only place where this message is used, so it can be removed from the i18n files.

JeroenDeDauw: Good catch

JeroenDeDauw: https://github.com/SemanticMediaWiki/SemanticResultFormats/commit/561384af15a33f5e16a235cf0f80d96fd04ed548

PR #445, opened 2018-10-17
Replace excel format by spreadsheet format

s7eph4n:
This PR is made in reference to: #382

This PR addresses or contains:
* Replace 'excel' format by 'spreadsheet' format
* Keep 'excel' as an alias
* Use new library [PhpSpreadsheet](https://github.com/PHPOffice/phpspreadsheet)
* Add fileformat parameter to #ask (Allowed values: xlsx, xls, ods, csv)

This PR includes:
- [ ] Update of RELEASE-NOTES.md
- [x] Tests (unit/integration)
- [ ] CI build passed

Fixes #382

PR #519, opened 2019-07-31
Make SRF installable without ext php-gd

JeroenDeDauw:
Refs: https://github.com/SemanticMediaWiki/SemanticResultFormats/pull/445

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -49,11 +49,11 @@
 		"mediawiki/semantic-media-wiki": "~3.0",
 		"nicmart/tree": "^0.2.7",
 		"data-values/geo": "~4.0|~3.0|~2.0",
-		"phpoffice/phpspreadsheet": "^1.4",
 		"symfony/css-selector": "^3.3",
 		"mediawiki/mermaid": "~2.1"
 	},
 	"suggest": {
+		"phpoffice/phpspreadsheet": "Required for 'format=spreadsheet'",
 		"phpoffice/phpexcel": "Required for 'format=excel'",
 		"mediawiki/graph-viz": "Required for 'format=graph' and 'format=process'"
 	},
--- a/SemanticResultFormats.php
+++ b/SemanticResultFormats.php
@@ -147,7 +147,6 @@ class SemanticResultFormats {
 			'dygraphs' => 'SRFDygraphs',
 			'incoming' => 'SRFIncoming',
 			'media' => 'SRF\MediaPlayer',
-			'spreadsheet' => 'SRF\SpreadsheetPrinter',
 			'datatables' => 'SRF\DataTables',
 			'gantt' => 'SRF\Gantt\GanttPrinter'
 		];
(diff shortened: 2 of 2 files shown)


composer.json
```diff
@@ -49,11 +49,11 @@
 		"mediawiki/semantic-media-wiki": "~3.0",
 		"nicmart/tree": "^0.2.7",
 		"data-values/geo": "~4.0|~3.0|~2.0",
-		"phpoffice/phpspreadsheet": "^1.4",
 		"symfony/css-selector": "^3.3",
 		"mediawiki/mermaid": "~2.1"
 	},
 	"suggest": {
+		"phpoffice/phpspreadsheet": "Required for 'format=spreadsheet'",
```
kghbln: This does not work since "spreadsheet" is enabled by default: https://github.com/SemanticMediaWiki/SemanticResultFormats/blob/master/DefaultSettings.php#L55

kghbln:
Why not having this by default? Probably in most cases you would like to have this.

composer.json
```diff
@@ -49,11 +49,11 @@
...
 		"phpoffice/phpexcel": "Required for 'format=excel'",
```
kghbln: This can probably go since "excel" is meant to be an alias of "spreadsheet", meaning that the old "phpexcel" lib is not needed in either case.

JeroenDeDauw:
I did not change working default behavior. If you have ext-gd installed the format will be enabled like before. The only behavior change is that if you do not have it installed, you can now install SRF.

Composer does not install PHP extensions.

kghbln:
> Composer does not install PHP extensions.

Ah, now I get it, "phpoffice/phpspreadsheet" somehow requires "php-gd"?!

JeroenDeDauw:
Yes it does. Which means that I cannot install SRF on my local dev wiki since I do not have this PHP extension.

kghbln:
> Yes it does. Which means that I cannot install SRF on my local dev wiki since I do not have this PHP extension.

Thanks for confirming!

mwjames:
The PR headline is misleading, you are removing
`phpoffice/phpspreadsheet` from being required which means users no
longer can expect that the `format=spreadsheet` works out of the box.
This needs to be noted in the RELEASE-NOTES and please check that the
`format=spreadsheet` is either disabled or outputs a message in case
`phpoffice/phpspreadsheet` is not installed.

As for the ext-gd, why is this an issue? I haven't heard any users
complaining that they cannot install SRF because package
`phpoffice/phpspreadsheet` requires ext-gd.

kghbln:
> As for the ext-gd, why is this an issue? I haven't heard any users complaining that they cannot install SRF because package `phpoffice/phpspreadsheet` requires ext-gd.

Well Jeroen is the first one to actually "complain" but also in the lucky situation that he can fix himself.

> The PR headline is misleading, you are removing `phpoffice/phpspreadsheet` from being required which means users no longer can expect that the `format=spreadsheet` works out of the box.

"spreadsheet" was not yet officially released. Since the the precursing "excel" was not available out of the box were is not breakage.

mwjames:
> "spreadsheet" was not yet officially released. Since the the precursing
> "excel" was not available out of the box were is not breakage.

Well, I don't think the result printer is checking whether assumed
phpspreadsheet classes are available or not since the package (via
Composer) is required which means either disable the format or check
in the class that `phpoffice/phpspreadsheet` is actually available
otherwise the result printer will blow up.

JeroenDeDauw:
@mwjames the format is not loaded when the PHP extension is not there

mwjames:
> @mwjames the format is not loaded when the PHP extension is not there

Okay then, if this has been confirmed allow Stephan (as the author) to
confirm this change.

s7eph4n:
I should probably have split #445 in two PRs, one for replacing excel, one for making it loaded by default.

Anyway, the situation now is as in 3.0 only that _excel_ is called _spreadsheet_ (with _excel_ being an alias).

PR #523, opened 2019-08-01
Remove unused mediawiki.legacy.wikibits dependency

Krinkle:
This provides addOnloadHook() and importScript/importStyles functions,
none of which are used here.


JeroenDeDauw:
Thank you!

PR #534, opened 2019-08-24
[ical] Add `DateParser`

mwjames:

- Isolates the date parsing and moves it into a separate class, adds a unit test, and fixes a copy-paste error `DataValueFactoryg::getInstance`

This PR includes:
- [x] CI build passed

### DIFF
--- a/src/iCalendar/DateParser.php
+++ b/src/iCalendar/DateParser.php
@@ -1,0 +1,64 @@
+<?php
+
+namespace SRF\iCalendar;
+
+use SMWDataValueFactory as DataValueFactory;
+use SMWTimeValue as TimeValue;
+
+/**
+ * @license GNU GPL v2+
+ * @since 3.1
+ */
+class DateParser {
+
+	/**
+	 * Extract a date string formatted for iCalendar from a SMWTimeValue object.
+	 *
+	 * @since 3.1
+	 *
+	 * @param TimeValue $dataValue
+	 * @param boolean $isEnd
+	 *
+	 * @return string
+	 */
+	public function parseDate( TimeValue $dataValue, $isEnd = false ) {
+		$year = $dataValue->getYear();
+
+		// ISO range is limited to four digits
+		if ( ( $year > 9999 ) || ( $year < -9998 ) ) {
+			return '';
+		}
+
+		$year = number_format( $year, 0, '.', '' );
+		$time = str_replace( ':', '', $dataValue->getTimeString( false ) );
+
+		// increment by one day, compute date to cover leap years etc.
+		if ( ( $time == false ) && ( $isEnd ) ) {
+			$dataValue = DataValueFactory::getInstance()->newDataValueByType(
+				'_dat',
+				$dataValue->getWikiValue() . 'T00:00:00-24:00'
+			);
+		}
+
+		$month = $dataValue->getMonth();
+
+		if ( strlen( $month ) == 1 ) {
+			$month = '0' . $month;
+		}
+
+		$day = $dataValue->getDay();
+
+		if ( strlen( $day ) == 1 ) {
+			$day = '0' . $day;
+		}
+
+		$result = $year . $month . $day;
+
+		if ( $time != false ) {
+			$result .= "T$time";
+		}
+
+		return $result;
+	}
+
+}
--- a/tests/phpunit/Unit/iCalendar/DateParserTest.php
+++ b/tests/phpunit/Unit/iCalendar/DateParserTest.php
@@ -1,0 +1,74 @@
+<?php
+
+namespace SRF\Tests\iCalendar;
+
+use SRF\iCalendar\DateParser;
+
+/**
+ * @covers \SRF\iCalendar\DateParser
+ * @group semantic-result-formats
+ *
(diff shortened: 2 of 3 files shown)


tests/phpunit/Unit/iCalendar/DateParserTest.php
```diff
@@ -0,0 +1,74 @@
...
+ */
+class DateParserTest extends \PHPUnit_Framework_TestCase {
+
+	public function testCanConstruct() {
+
+		$this->assertInstanceOf(
+			DateParser::class,
+			new DateParser()
+		);
+	}
```
JeroenDeDauw: I think this test is better omitted

tests/phpunit/Unit/iCalendar/DateParserTest.php
```diff
@@ -0,0 +1,74 @@
...
+
+	public function testParseDate_Year() {
+
+		$timeValue = $this->getMockBuilder( '\SMWTimeValue' )
+			->disableOriginalConstructor()
+			->getMock();
```
JeroenDeDauw: Pretty sure you can use createMock rather than this deprecated form

You can also use SMWTimeValue::class

tests/phpunit/Unit/iCalendar/DateParserTest.php
```diff
@@ -0,0 +1,74 @@
...
+
+		$timeValue->expects( $this->any() )
```
JeroenDeDauw: SMWTimeValue is a value object right? Then why are you creating a mock? This adds a bunch of hidden coupling (method names as strings, plus their signatures...) which is best avoided if there is no good reason for it.

tests/phpunit/Unit/iCalendar/DateParserTest.php
```diff
@@ -0,0 +1,74 @@
...
+namespace SRF\Tests\iCalendar;
+
+use SRF\iCalendar\DateParser;
+
+/**
+ * @covers \SRF\iCalendar\DateParser
+ * @group semantic-result-formats
+ *
+ * @license GNU GPL v2+
+ * @since 3.1
```
JeroenDeDauw: This class is not package public so this is clutter IMO

src/iCalendar/DateParser.php
```diff
@@ -0,0 +1,64 @@
...
+	 * Extract a date string formatted for iCalendar from a SMWTimeValue object.
+	 *
+	 * @since 3.1
+	 *
+	 * @param TimeValue $dataValue
+	 * @param boolean $isEnd
+	 *
+	 * @return string
+	 */
+	public function parseDate( TimeValue $dataValue, $isEnd = false ) {
```
JeroenDeDauw: public function parseDate( TimeValue $dataValue, bool $isEnd = false ): string {

tests/phpunit/Unit/iCalendar/DateParserTest.php
mwjames: I have testing the constructor for a while now and it served me well,it  even found some issues in other tests.

tests/phpunit/Unit/iCalendar/DateParserTest.php
mwjames: I know but the intent is clearer that constructor is disabled, and we don't count lines so it doesn't really matter.

tests/phpunit/Unit/iCalendar/DateParserTest.php
mwjames: > SMWTimeValue is a value object right? 

Sort of but using an a mock ensures that specific coupling to let's say a Title object, or the Store are not pulled in hereby avoids creating a possible DB connection, especially the Title object is nasty, and by using a mock I'm certain I have an encapsulate object that doesn't create any connection.

tests/phpunit/Unit/iCalendar/DateParserTest.php
mwjames: I'm not a fan of the "no comment" purist approach, so do I think some minimal style block should be retained.

src/iCalendar/DateParser.php
mwjames: Doing strict typing, I know you like it but I have not come to terms on to how to approach this without creating a piece here and there. Without a clear map of how to transform an entire "package" (it would refer here to the entire ical format and its depending classes) and not just a one class I'm a bit reluctant to added only once.

tests/phpunit/Unit/iCalendar/DateParserTest.php
JeroenDeDauw: Eh... both the authors of PHPUnit and me beg to differ

tests/phpunit/Unit/iCalendar/DateParserTest.php
JeroenDeDauw: Alright, sounds like SMWTimeValue is not a value object then.

tests/phpunit/Unit/iCalendar/DateParserTest.php
JeroenDeDauw: I was talking just about the @since tag

PR #536, opened 2019-08-24
Use `sequence_map` for author/bibtex test

mwjames:
This PR is made in reference to: https://github.com/SemanticMediaWiki/SemanticMediaWiki/issues/4226

- This is an example for the use of a "sequence map", while MySQL/MariaDB, and SQLite do return a specific sequence by default, PostgreSQL would fail that assumption and the test as well
- After adding a `sequence_map` for the `Has author` property the expected author sequence is produced on all platforms equally

This PR includes:
- [x] CI build passed

## Note

```
1) SRF\Tests\Integration\JSONScript\JsonTestCaseScriptRunnerTest::testCaseFile with data set "bibtex-01.json" ('/home/travis/build/SemanticMe...1.json')
Failed "#2 `format=bibtex` multiple authors (bibtex-01-2.bib)" for StringContains:
==== (actual) ====
@Book{stegun1964homf,
  address = "New York",
  author = "Irene A. Stegun and Milton Abramowitz",
  edition = "ninth Dover printing, tenth GPO printing",
  publisher = "Dover",
  title = "Handbook of Mathematical Functions",
  year = "1964",
}
==== (expected) ====
[ @Book{abramowitz1964homf,
  address = "New York",
  author = "Milton Abramowitz and Irene A. Stegun",
  edition = "ninth Dover printing, tenth GPO printing",
  publisher = "Dover",
  title = "Handbook of Mathematical Functions",
  year = "1964",
} ]
```

### DIFF
--- a/tests/phpunit/Integration/JSONScript/Fixtures/bibtex-sequence-map.json
+++ b/tests/phpunit/Integration/JSONScript/Fixtures/bibtex-sequence-map.json
@@ -1,0 +1,9 @@
+{
+    "type": "PROPERTY_PROFILE_SCHEMA",
+    "profile": {
+        "sequence_map": true
+    },
+    "tags": [
+        "property profile"
+    ]
+}
--- a/tests/phpunit/Integration/JSONScript/TestCases/bibtex-01.json
+++ b/tests/phpunit/Integration/JSONScript/TestCases/bibtex-01.json
@@ -2,9 +2,16 @@
 	"description": "Test `format=bibtex`",
 	"setup": [
 		{
+			"namespace": "SMW_NS_SCHEMA",
+			"page": "Profile:AuthorSequence",
+			"contents": {
+				"import-from": "/../Fixtures/bibtex-sequence-map.json"
+			}
+		},
+		{
 			"page": "Has author",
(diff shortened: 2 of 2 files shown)


mwjames:
@kghbln Some anecdotal notes, the failing bibtex test actually was the reason I added the "sequence_map" in SMW knowing that some properties require a specific order of their values. As outlined above, depending on which system you create the result (or export the content) you could get different results unless you define something what we now call a "sequence map".

kghbln:
Thanks for the note!

mwjames:
> Thanks for the note! Added to wiki.

To clarify, the PR only adds the "sequence_map" to `Has author` property. Nothing changes for the bibtex format itself, only the integration test is modified to use the new SMW core feature to ensure that the export output is consistent on all platforms given that we annotate `[[Has author::Milton Abramowitz]] [[Has author::Irene A. Stegun]]` hereby expecting the export to contain ` author = "Milton Abramowitz and Irene A. Stegun"`.

kghbln:
Yes, but before it would not have been possible to use the sequence map feature for this result format. At least this is what I meant with my note on the wiki.

PR #538, opened 2019-08-26
Create GraphFormatter Class

RacoonDev:
This PR is made in reference to: #297 

- GraphFormatter Class
- php unit tests for GraphFormatter
- code improvements

This PR includes:
- [ ] CI build passed

At the moment the test case for the buildGraph() is not covered but will be added in the next few days.


i18n/ar.json
```diff
@@ -170,7 +170,6 @@
 	"srf_paramdesc_hashname": "إذا تم توفيرها وإمكانية تمديد جداول التجزئة فسيؤدي هذا إلى إنشاء مزيج بالاسم المحدد (لا يوجد إخراج مرئي بعدئذ)",
 	"srf-printername-graph": "رسم بياني",
 	"srf-paramdesc-graph-relation": "يحدد ما إذا كانت الموضوعات أو خصائص الاسم آباء أم أطفال",
-	"srf-paramdesc-graph-nameprop": "لتعيين الخاصية التي سيتم استخدامها كموضوع بدلا من الموضوع الفعلي: أي اسم الصفحة",
```
JeroenDeDauw: Try doing one thing per commit please. This could have been in a dedicated PR

src/Graph/GraphFormatter.php
```diff
@@ -0,0 +1,229 @@
...
+
+use Html;
+
+/**
+ *
+ *
+ * @see https://www.semantic-mediawiki.org/wiki/Help:Graph_format
+ *
+ * @license GNU GPL v2+
+ * @since 3.1
```
JeroenDeDauw: Either 3.2 or just drop it

src/Graph/GraphFormatter.php
```diff
@@ -0,0 +1,229 @@
...
+	private function add( $line ){
+		$this->graph .= $line;
+	}
+
+
+	/*
+	* Creates the DOT (graph description language) which can be processed by the graphviz lib
+	*
+	* @see https://www.graphviz.org/
+	* @since 3.1
```
JeroenDeDauw: same

src/Graph/GraphFormatter.php
```diff
@@ -0,0 +1,229 @@
...
+	*
+	@param SRF\Graph\GraphNodes[] $nodes
```
JeroenDeDauw: bad formatting

tests/phpunit/Unit/Graph/GraphFormatterTest.php
```diff
@@ -0,0 +1,102 @@
...
+use SRF\Graph\GraphFormatter;
+use SRF\Graph\GraphNode;
+
+
+/**
+ * @covers \SRF\Graph\GraphFormatter
+ * @group semantic-result-formats
+ *
+ * @license GNU GPL v2+
+ * @since 3.1
```
JeroenDeDauw: wrong, just drop the since tag

tests/phpunit/Unit/Graph/GraphFormatterTest.php
```diff
@@ -0,0 +1,102 @@
...
+
+
+		print('------expected-------');
+		print($expected);
+		print('-------------');
+
+
+		$this->assertEquals( $this->graphFormatter->getGraph(), $expected );
+	}
+	*/
```
JeroenDeDauw: Please remove

src/Graph/GraphFormatter.php
```diff
@@ -0,0 +1,229 @@
...
+
+	/**
+	 * Returns the word wrapped version of the provided text.
+	 *
+	 * @param string $text
+	 * @param integer $charLimit
+	 *
+	 * @return string
+	 */
+	public static function getWordWrappedText( $text, $charLimit ) {
```
JeroenDeDauw: Did you make this public and static so you can test it?

JeroenDeDauw: Also, you can do

```php
getWordWrappedText( string $text, int $charLimit ): string {
```

and then drop the whole docblock

src/Graph/GraphFormatter.php
RacoonDev: @see: https://github.com/RacoonDev/SemanticResultFormats/commit/10eb56e4c1a888dd4f9988427e55bf3f12f98917

src/Graph/GraphFormatter.php
RacoonDev: @see: https://github.com/RacoonDev/SemanticResultFormats/commit/10eb56e4c1a888dd4f9988427e55bf3f12f98917

tests/phpunit/Unit/Graph/GraphFormatterTest.php
RacoonDev: @see: https://github.com/RacoonDev/SemanticResultFormats/commit/10eb56e4c1a888dd4f9988427e55bf3f12f98917

src/Graph/GraphFormatter.php
RacoonDev: No I think that "public static" is best for this function because it is not bound to a GraphFormatter object.

tests/phpunit/Unit/Graph/GraphFormatterTest.php
RacoonDev: @see: https://github.com/RacoonDev/SemanticResultFormats/commit/10eb56e4c1a888dd4f9988427e55bf3f12f98917

src/Graph/GraphFormatter.php
RacoonDev: @see: https://github.com/RacoonDev/SemanticResultFormats/commit/10eb56e4c1a888dd4f9988427e55bf3f12f98917

i18n/ar.json
RacoonDev: So should i revert this commit?

JeroenDeDauw: No, just an FYI

src/Graph/GraphFormatter.php
JeroenDeDauw: Maybe it should be its own class? (I'm not sure since I have not looked at this long enough)

Global functions are generally bad, and static functions are global functions. Then again this one is not accessing any state, and pure function leaf nodes can be global without too much issues.

src/Graph/GraphFormatter.php
```diff
@@ -0,0 +1,227 @@
...
+		'darkgreen',
+		'yellow',
+		'darkblue',
+		'magenta',
+		'steelblue2'
+	];
+	private $legendItem = [];
+	private $options = [];
+
+	public function __construct( $options ){
```
JeroenDeDauw: An object to hold the options rather than an array has the big advantage that you can use code analysis (things like auto completion, automated refactoring and type mismatch detection)

tests/phpunit/Unit/Graph/GraphFormatterTest.php
```diff
@@ -0,0 +1,96 @@
...
+		$this->nodes[] = $node1;
+
+		$node2 = new GraphNode( 'Team:Beta' );
+		$node2->setLabel( "Beta" );
+		$node2->addParentNode( "Casted", "Person:Sebastian Schmid" );
+		$node2->addParentNode( "Casted", "Person:Alexander Gesinn" );
+		$node2->addParentNode( "Part of Team ", "Team:Alpha" );
+		$this->nodes[] = $node2;
+
+		$this->graphFormatter->buildGraph( $this->nodes );
```
JeroenDeDauw: Looks like there is no need to have class level state here. You could just have a newGarphFormatter method that you call when you need it.

JeroenDeDauw:
Would be good to tackle the options array vs object thing since that is a maintainability regression. Apart from that the code looks OK enough to me.

src/Graph/GraphFormatter.php
RacoonDev: @see: https://github.com/RacoonDev/SemanticResultFormats/commit/ade773b60f5bdb9dabfe95d21b30251298898c90

JeroenDeDauw: Huh.

Sorry it seems I was not clear enough. My concern is not about object vs array SYNTAX. It is about free-form-dynamic-whatever vs defined-structure. In other words, I was thinking about a class like GraphFormatterOptions with fields similar to those that where present in the class you are extracting from.

JeroenDeDauw: Just a bunch of public fields in this class will already be a big improvement. A next step could be adding type docs to the fields.

RacoonDev: Okey I will create this class but I think "GraphOptions" would be a better name for it. Should I just use public fields or should I declare them private and implement some getter functions?

JeroenDeDauw: PHP does not have a nice way to create this type of objects... you can read my rant on it here https://www.entropywins.wtf/blog/2016/02/03/missing-in-php7-value-objects/

Since this `GraphOptions` is used only locally I'd be tempted with just having public fields and saving myself the hassle of adding extra stuff.

RacoonDev: I created now the GraphOptions Class with its public fields and the related getter functions. For the initialising i pass the params-array of the SMWResultPrinter to the GraphOptions constructor and map the public fields based on the array keys.

JeroenDeDauw: i see you went with private fields and getters. This works quite nicely since you need to convert from an array anyway, meaning your constructor does not end up with 1000 parameters. This is nicer than what I was thinking about. You now also have return types \o/ That means the type docs for the fields are pretty superfluous and can be removed.

src/Graph/GraphPrinter.php
```diff
@@ -298,7 +134,7 @@ protected function processResultRow( array /* of SMWResultArray */ $row ) {
 				// create SRF\GraphNode for column 0
 				if ( $i == 0 ) {
 					$node = new GraphNode( $object->getShortWikiText() );
-					$node->setLabel($object->getDisplayTitle());
+					$node->setLabel( $object->getDisplayTitle() );
```
mwjames: @RacoonDev `getDisplayTitle` is not guaranteed to have a string representation except for when a display title is available (as the name suggests), meaning that if a page has no display title (aka DISPLAYTILE) then this function will return with an empty string therefore you should rely on `getPreferredCaption` instead to have some title representation.

RacoonDev: @mwjames thanks for the advice, I've replaced the functions. I think there is no reason to overwrite the "getPreferredCaption" but please tell me if I'm wrong.

Issue #549, opened 2019-10-23
Graph format: $node variable is not initialized in the else block


jewloszewicz:
### Setup

- MW version: 1.31.5
- SMW version: 3.1.0
- SRF version: 3.2.0-alpha

### Issue

Error from line 186 of /var/www/smw/extensions/SemanticResultFormats/src/Graph/GraphPrinter.php: Call to a member function addParentNode() on null

JeroenDeDauw:
Thanks for reporting this issue!

Ping @SbstnS

SbstnS:
@jewloszewicz can you please post your inline query which causes this error so i can reproduce it.

SbstnS:
@jewloszewicz I'm working with the same setup as you've described above but I never run into this error. 

You may have made some mistakes during installing the extension.
Please check the installation guide.

https://www.semantic-mediawiki.org/wiki/Extension:Semantic_Result_Formats/Installation

jewloszewicz:
@SbstnS Unfortunately, I don't have access to that instance anymore.

There is mistake in the code. $node in the else statement is never initialized.

https://github.com/SemanticMediaWiki/SemanticResultFormats/blob/1b983766322732e50fa6f613e006c223f7bd9533/src/Graph/GraphPrinter.php#L183

SbstnS:
@jewloszewicz I'm not sure how then you can run into this error that you've described above....

`Error from line 186 of /var/www/smw/extensions/SemanticResultFormats/src/Graph/GraphPrinter.php: Call to a member function addParentNode() on null`

To me everything works as expected! 
If i had your inline query than i could cover the hole test scenario...

Here is the one that I used for testing.

```
{{#ask:[[Category:Team]]
  |?Part Of Team
  |format=graph
  |graphcolor=yes
  |graphlink=yes
  |graphname=PeopleKnows
  |graphlegend=yes
  |graphlabel=yes
  |graphsize=10,100
  |nodelabel=displaytitle
}}
```

SbstnS:
@JeroenDeDauw 
But I noticed a few parts while testing the code that could be improved. I'll do this in a new PullRequest.

alex-mashin:
I think, this bug has benn fixed by  #749.

PR #550, opened 2019-11-08
Add additional check if $node has already been initialized

SbstnS:
This PR is made in reference to: #549

### DIFF
--- a/src/Graph/GraphPrinter.php
+++ b/src/Graph/GraphPrinter.php
@@ -171,16 +171,18 @@ class GraphPrinter extends ResultPrinter {
 		foreach ( $row as $i => $resultArray ) {
 
 			// loop through all values of a multivalue field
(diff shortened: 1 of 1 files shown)


SbstnS:
Travis failed because of error in Outline format 

> 1) SRF\Tests\Outline\OutlineResultPrinterTest::testGetResult_LinkOnNonFileOutput
TypeError: Return value of Parser::getTitle() must be an instance of Title, null returned

Also the Mermaid tests were skipped because missing config for php 7.3. 

> Dependency: Mermaid as requirement is not available!

@mwjames do we need to add an new test case for php 7.3 to the Mermaid extension?
https://github.com/SemanticMediaWiki/Mermaid/blob/master/.travis.yml

gesinn-it:
@SbstnS @JeroenDeDauw This PR is obsolete and can be canceled / deleted.

PR #563, opened 2019-12-31
Replace deprecated DB_SLAVE constant

Ammarpad:
This constant was renamed to DB_REPLICA and remained as deprecated alias since MW 1.28.

It was completely [dropped in MW 1.34](https://gerrit.wikimedia.org/r/#/c/mediawiki/core/+/443747/)

### DIFF
--- a/formats/Exhibit/SRF_Exhibit.php
+++ b/formats/Exhibit/SRF_Exhibit.php
@@ -71,7 +71,7 @@ class SRFExhibit extends SMWResultPrinter {
 			$remote = true;
 
 			// fetch interwiki link
-			$dbr = &wfGetDB( DB_SLAVE );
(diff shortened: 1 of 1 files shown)


kghbln:
This change takes MW 1.31 to 1.33 into account? In other words: I think we already had this in the code and it blew up everything.

JeroenDeDauw:
@Ammarpad said this is in MW since 1.28, in which case it is no issue in this extension since it requires 1.31 anyway.

kghbln:
Assuming good faith. :)

JeroenDeDauw:
Good fluff

Fluff faith

Fluff fluff

kghbln:
Well, it should be fluff otherwise we have grump. Do not like grump.

JeroenDeDauw:
> Do not like grump.

  `[citation needed]`

PR #582, opened 2020-03-30
Add intro-/outrotemplate

### LINKED ISSUE #578: [Outline] Support introtemplate and outrotemplate parameter
### Issue
At the moment, Outline provides a template mechanism, that allows to specify templates for segment header and items.

As a user I'd like to be able to specify an intro and outro template that surrounds the complete Outline's results. This could for example be a container `<div>` tag.

I propose, on the basis of the existing template/plainlist format parameters, to use `introtemplate` and `outrotemplate` as parameter names.

@kghbln @s7eph4n @JeroenDeDauw @mwjames : what do you think?

SbstnS:
This PR is made in reference to: #578

This PR contains:
- add intro-/outrotempalte for outline format

Fixes #578

### DIFF
--- a/src/Outline/OutlineResultPrinter.php
+++ b/src/Outline/OutlineResultPrinter.php
@@ -123,20 +123,13 @@ class OutlineResultPrinter extends ResultPrinter {
 			$result .= $link->getText( $outputMode, $this->mLinker ) . "\n";
 		}
 
-
-		// Add introtemplate
-		if ( $this->params['introtemplate'] !== '' ) {
-			$introtemplate = "{{" . $this->params['introtemplate'] . "}}";
-
-			$result = $introtemplate . $result;
+		if( $this->params['introtemplate'] !== '' || $this->params['introtemplate'] !== ''){
+			$this->hasTemplates = true;
 		}
 
-		// Add outrotemplate
-		if ( $this->params['outrotemplate'] !== '' ) {
-			$outrotemplate = "{{" . $this->params['outrotemplate'] . "}}";
+		$result = $this->getIntroTemplate() . $result;
 
-			$result .= $outrotemplate;
-		}
+		$result .= $this->getOutroTemplate();
 
 		return $result;
 	}
@@ -163,4 +156,17 @@ class OutlineResultPrinter extends ResultPrinter {
 		return $outlineItem;
 	}
 
+	function getIntroTemplate(): string {
+		if ( $this->params['introtemplate'] !== '' ) {
+			return "{{" . $this->params['introtemplate'] . "}}";
+		}
+		return "";
+	}
+
+	function getOutroTemplate(): string {
+		if ( $this->params['outrotemplate'] !== '' ) {
+			return "{{" . $this->params['outrotemplate'] . "}}";
+		}
+		return "";
+	}
 }
--- a/src/Outline/OutlineResultPrinter.php
+++ b/src/Outline/OutlineResultPrinter.php
@@ -123,14 +123,6 @@ class OutlineResultPrinter extends ResultPrinter {
 			$result .= $link->getText( $outputMode, $this->mLinker ) . "\n";
 		}
 
-		if( $this->params['introtemplate'] !== '' || $this->params['introtemplate'] !== ''){
-			$this->hasTemplates = true;
-		}
-
-		$result = $this->getIntroTemplate() . $result;
-
-		$result .= $this->getOutroTemplate();
-
 		return $result;
 	}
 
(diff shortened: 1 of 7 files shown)


src/Outline/OutlineResultPrinter.php
```diff
@@ -111,6 +123,21 @@ protected function getResultText( QueryResult $res, $outputMode ) {
 			$result .= $link->getText( $outputMode, $this->mLinker ) . "\n";
 		}
 
+
+		// Add introtemplate
```
JeroenDeDauw: Comments like this have downsides, see something like Clean Code by Robert C Martin.

src/Outline/OutlineResultPrinter.php
```diff
@@ -111,6 +123,21 @@ protected function getResultText( QueryResult $res, $outputMode ) {
...
+		if ( $this->params['introtemplate'] !== '' ) {
+			$introtemplate = "{{" . $this->params['introtemplate'] . "}}";
+
+			$result = $introtemplate . $result;
+		}
```
JeroenDeDauw: You can avoid the comment, and a bunch of complexity, by creating a new function. 

`$result = $this->getIntroTemplate() . $result;`

```php
function getIntroTemplate(): string {
    // return empty string if param not set

    // return "{{" . $this->params['introtemplate'] . "}}"
}
```

SbstnS: ok thanks for your review I will remove the comments and put the functionality to new functions.

SbstnS:
I will also include integration tests for the new parameters.

mwjames:
This function shouldn't be placed here, there is a `TemplateBuilder`
which encapsulates template code handling.

JeroenDeDauw:
@mwjames just had a look at that class. You might be misremembering what you did there. It takes and OutlineTree object and is not a generic template builder. Does not look applicable to this PR.

mwjames:
> @mwjames just had a look at that class. You might be misremembering what you
> did there. It takes and OutlineTree object and is not a generic template
> builder. Does not look applicable to this PR.

`templatebuilder->build` returns a string with the template code and
any introtemplate or outrotemplate should only be applied to that
result not any other code path (which is what is proposed here) and
the reason those additional parameters are to be added as art of
`TemplateBuilder::build`.

SbstnS:
@mwjames, @JeroenDeDauw  unfortunately, the discussion has passed me by and i made some commits for this SRF format. If I understand this correctly, the intro and outro template should only work in combination with the "template" parameter?

src/Outline/OutlineResultPrinter.php
```diff
@@ -136,4 +156,17 @@ private function newOutlineItem( $row ) {
 		return $outlineItem;
 	}
 
+	function getIntroTemplate(): string {
+		if ( $this->params['introtemplate'] !== '' ) {
```
JeroenDeDauw: Would have been ever so slightly simpler if you did not do the negation here.

tests/phpunit/Unit/Outline/OutlineResultPrinterTest.php
```diff
@@ -82,9 +82,37 @@ public function testGetResult_LinkOnNonFileOutput() {
...
+			->getMock();
+
+		$outrotemplate->expects( $this->any() )
+			->method( 'getName' )
+			->will( $this->returnValue( 'outrotemplate' ) );
+
+		$outrotemplate->expects( $this->any() )
+			->method( 'getValue' )
+			->will( $this->returnValue( '' ) );
+
```
JeroenDeDauw: Oh man this is an unfortunate pattern that you copied here. Thanks for writing the test though!

SbstnS:
@mwjames @JeroenDeDauw would be great if you could have a look at my code changes.

src/Outline/TemplateBuilder.php
```diff
@@ -153,4 +153,18 @@ private function close() {
...
 
+	function getIntroTemplate(): string {
+		if ( $this->params['introtemplate'] ) {
+			return $this->open( $this->params['introtemplate'] ) . $this->close();
+		}
+		return "";
+	}
+
+	function getOutroTemplate(): string {
+		if ( $this->params['outrotemplate'] ) {
```
JeroenDeDauw, line 164: Oops, I should have been more clear. With dropping the negation I meant changing `!==` to `===`. Explicit checks are definitely better than implicit checks. Right now if you have a template named `0` the check will fail.

JeroenDeDauw: https://github.com/SemanticMediaWiki/SemanticResultFormats/pull/585

Issue #584, opened 2020-04-01
[Eventcalendar] Including end date for multi-day events


mint-thompson:
### Setup

- MW version: 1.32.1
- DB (MySQL etc.): MySQL 5.6.47
- PHP version: 7.2.29
- SMW version: 3.0.1
- SRF version: 3.0.2-alpha (43a8844)
- Browsers and versions (if applicable): Tested in Chrome 80, IE 11, Edge 44

### Issue

When displaying events that span multiple days, the eventcalendar format shows the event on each day that it occurs. However, this can be troublesome when the start and end of the event are stored as dates without times. In this case, a default time of midnight is used, and the end date is not included in the range. For example, a start date of `2020 March 23` and end date of `2020 March 27` will occupy Monday (March 23), Tuesday (March 24), Wednesday (March 25), and Thursday (March 26), but not Friday (March 27).

**Steps to reproduce**

The sandbox is currently refusing to let me confirm my email address, so I can't build a demonstration there. I will edit this issue to include a link to it as soon as that is resolved.

I have a solution ready to go in [my fork](https://github.com/mint-thompson/SemanticResultFormats/tree/calendar-end-date), but I will try to get the demonstration on the sandbox first.

kghbln:
Yeah, handling events and recurring events can indeed require some voodoo. If there is something to improve the cause ...!

mint-thompson:
The sandbox is still insisting that my email is unconfirmed and that I've tried to confirm my email too many times, so I went ahead and created the PR.

kghbln:
Thanks for the pull! I guess it is just a matter of clicking on the link you got via e-mail. Either a typo in the e-mail or deep in your spam filter. The wiki uses Amazon SES so it should be fairly reliable.

mwjames:
> the range. For example, a start date of 2020 March 23 and end date of 2020 March 27 will occupy Monday (March 23), Tuesday (March 24), Wednesday (March 25), and Thursday (March 26), but not Friday (March 27).

@mint-thompson Just to clarify, there isn't any other why to deduce from existing information that the end-date has to be included and requires an extra parameter to come to that conclusion?

mint-thompson:
[...]

PR #586, opened 2020-04-01
[Eventcalendar] Including end date for multi-day events

mint-thompson:
This PR is made in reference to: #584

- new parameter `includeend` for the eventcalendar format
- messages in `qqq.json` and `en.json` to describe this parameter

* Support parameter "value filter pagedefaultsort", which can be specified to indicate that the expected value for the field is the name of a page, and that page's defaultsort attribute should be used to order the filter checkboxes rendered for that field.

* More meaningful variable names

* Add includeend parameter for eventcalendar format

When the includeend parameter is set to true, the end date is
included in the date range for an event. This is useful on wikis
where event start and end dates, but not times, are entered. Since
the default time is midnight, the entered end date would normally
not be included in the date range.

The desired output is achieved by adding one day to the end date
before setting the rowData. This is similar to how the eventcalendar
library implements its nextDayThreshold functionality.

* includeend parameter defaults to true

Based on a suggestion from akuckartz, this parameter has a default
value of true, as this is more likely to be the desired behavior.

Added a comment explaining the precision bitmask being checked from
smw.dataItem.time.

This PR includes:
- [ ] CI build passed


akuckartz:
I think that the end date should normally be included. Maybe an optional configuration parameter should be provided for some time allowing to actively configure the previous semantics?

formats/calendar/resources/ext.srf.formats.eventcalendar.js
```diff
@@ -201,7 +201,11 @@
 											dates.push( value.getMwTimestamp() );
 											rowData.allDay = true;
 										} else {
-											rowData.end = value.getDate().toISOString();
+											var dataDate = value.getDate();
+											if(data.query.ask.parameters.includeend && ( value.precision & 8 ) == 0 ) {
+												dataDate.setDate(dataDate.getDate() + 1);
```
JeroenDeDauw, line 206: `( value.precision & 8 ) == 0`? huh? bit magic that does something... I'm sure this can be expressed more clearly

mint-thompson: `precision` is a bitmask set in [ext.smw.dataItem.time.js](https://github.com/SemanticMediaWiki/SemanticMediaWiki/blob/master/res/smw/data/ext.smw.dataItem.time.js#L35), so I'll add a comment here giving a brief explanation of the bitmask.

JeroenDeDauw:
Needs a rebase against master

mwjames:
It still don't think the extra parameter is necessary and the information can be deduced from existing information but I just don't have the time to go through the code therefore instead have collecting rust on this PR, will merge this as-is.

kghbln:
I did not really document the new parameter however I added an [information about it](https://www.semantic-mediawiki.org/wiki/Help:Eventcalendar_format#Format_specific) pointing to this very pull allowing for easier discovery.

Issue #591, opened 2020-04-20
Filtered format does not use available space for placeholder text


krabina:
### Setup

MediaWiki | 1.31.7
-- | --
PHP | 7.2.29 (litespeed)
MariaDB | 10.3.22-MariaDB
ICU | 50.2
SMW | 3.1.5
SRF | 3.1.0 (aa0649f) 17:55, 18. Aug. 2019

### Issue
In filter select field of type tokens, the filtered format shows a placeholder text ("select a filter value"). The available space is not used, the wording (in the example in German) is cropped after 27 characters or so:

![grafik](https://user-images.githubusercontent.com/4318745/79729255-81851a80-82ef-11ea-9ffc-522c89962ecc.png)

If you remove this from the css, the problem goes away:

```
.filtered-filters .filtered-value .select2-search__field {
    width: 100% !important;
}
```

### desired result
![grafik](https://user-images.githubusercontent.com/4318745/79730183-bf367300-82f0-11ea-9bd7-a70f650f659c.png)

krabina:
Interestingly, in SRF 3.0.1, removing the CSS does the opposite: in an (english version) wiki: removing the above results in:
![grafik](https://user-images.githubusercontent.com/4318745/79730819-af6b5e80-82f1-11ea-84ff-1cea22e7157c.png)
In 3.1.0 it does work, though

krabina:
was a configuration problem

PR #592, opened 2020-04-20
Update ext.srf.filtered.value-filter.less

krabina:
seems to be not necessary in SRF 3.1.x. On the contrary, it crops the paceholder text (can be seen in German language settings).

This PR is made in reference to: #591

This PR needs some more testing. Tested it on my wikis, but need someone to confirm nothing breaks with this PR.

This PR includes:
- [ ] CI build passed

Fixes #


krabina:
sorry, this does not work

PR #596, opened 2020-05-01
Replace use of deprecated $wgParser

cscott:
Since the minimum MW version is already set to 1.31, we can remove some
ancient back-compat code, and we know that MediaWikiServices::getParser()
will always be available.

Bug: T160811


kghbln:
Thanks for the pull!

cscott:
This is also https://phabricator.wikimedia.org/T241555; I didn't notice that a more-specific bug had been filed for this.

PR #598, opened 2020-06-04
Add {{#itemsubjectraw}} for furhter processing

SbstnS:
- a new variable {{#itemsubjectraw}}

This PR includes:
- [ ] Update of README.md fo Outline format
- [ ] Update of TemplateBuilder.php

Currently the Outline format does support

1. linking of subject items (which automatically makes use of Display Title) or 
2. unlinked subject items, showing the Display Title

If you want to use the template format in Outline you might want to look up additional fields. For this, you have to use option 2. because linked items can not be used in "ask queries". Unfortunately, option 2. does only return the Display Title, not the page title.

To solve this, this PR introduces an additional template parameter #itemsubjectraw which can be used inside the item's template for further processing.

### DIFF
--- a/src/Outline/README.md
+++ b/src/Outline/README.md
@@ -27,6 +27,7 @@ The `...-item` template will also provide additional information such as:
 
 - `#itemsection` section number of the item
 - `#itemsubject` the subject (aka page) of the item processed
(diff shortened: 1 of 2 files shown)


JeroenDeDauw:
Please also update RELEASE-NOTES.md

PR #604, opened 2020-07-29
adding statistical result formats

krabina:
This PR addresses or contains: more math/statistical functions
- variance and samplevariance
- standarddeviation and samplestandarddeviation
- median
- mode
- quartilupper and quartilupper.exc
- quartillower and quartillower.exc
- interquartilerange and interquartilerange.exc
- interquartilemean

you can see it [here](https://testwiki2.kdz.eu/index.php?title=Math_formats#New_formats)
documentation of these new result formats will be provided as soon as the PR is accepted into master

(The calculations of variance and standarddeviation vary if the ask query includes only a sample of the data or all of the data, therefore there are two formats. The same goes for the calculations where in Excel you have ein .inc and .exc formula: here quartilupper = quartile.incl in Excel and quartilupper.exc = quartil.exc in ExcelI)

tested on

MediaWiki | 1.31.7
PHP | 7.2.32 (fpm-fcgi)
MariaDB | 10.3.23-MariaDB
ICU | 50.2
SMW 3.1.6

This PR includes:
- [ ] CI build passed


JeroenDeDauw:
The added code does not depend on SMW and could be put into dedicated pure functions outside of the result printer. This would also make it trivial to write some simple tests for them.

krabina:
What do you mean "does not depend on SMW". It takes values from an ask query and does calculations with it. We just extended what was there in SRF_Math.php to have more functions in the same manner.

Changing SRF_Math.php in general might be a different task outside of the scope of this PR.

JeroenDeDauw:
The code can be in functions that take the array of numbers and return a number. Those functions do not depend on SMW or MediaWiki, making testing very easy.

I'm not suggesting to improve the existing code. Would be nice to not make the mess worse, which can be done via dedicated functions. The main benefit being increased confidence that the code works as expected.

krabina:
I see. So instead of 
```
 case 'average':
	return array_sum( $numbers ) / count( $numbers );
	break;
```
you would like to see

```
 case 'average':
	return average( $numbers ) ;
	break;
```

and at the bottom a 
`private function average`
that does the actual calulcation?

JeroenDeDauw:
Yeah - basically. Though to test these you could make them `public static` functions. And perhaps best to stuff them in a dedicated "class" rather than the result printer.

JeroenDeDauw:
So

```php
class MathStuffs {
    public static function average( array $numbers ): float {
        return array_sum( $numbers ) / count( $numbers );
    }
}
```

krabina:
like this?

krabina:
This should be ready now from our side...

krabina:
Median was already part of SRF bevore, btw.

PR #607, opened 2020-08-06
Remove check for MediaWiki environment from settings file

t-tomalak:
Currently, SemanticResultsFormats autoloads its setup file via Composer's autoloader and exits if it does not detect a MediaWiki environment. This prevents developers from executing tools also installed via composer such as the `phpunit` binary, or `phpcs`/`phpcbf`/`phan` etc., as SemanticResultFormat will immediately exit the process. As a fix, remove the environment check; this is also consistent with how Semantic MediaWiki itself behaves since SemanticMediaWiki/SemanticMediaWiki#1732 and SemanticMediaWiki/Mermaid#47.

This PR includes:
- [ ] CI build passed

### DIFF
--- a/DefaultSettings.php
+++ b/DefaultSettings.php
@@ -9,10 +9,6 @@
  * @author Daniel Werner < danweetz@web.de >
  */
 
-if ( !defined( 'MEDIAWIKI' ) ) {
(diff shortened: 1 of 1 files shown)


JeroenDeDauw:
@t-tomalak this is released now

PR #610, opened 2020-08-19
Remove and replace usages of Skin::makeVariablesScript()

Ammarpad:
The function Skin::makeVariablesScript() will be deprecated in MediaWiki 1.36. Since it's used in a lot of places in this codebase, I reimplemented it in SRF utils class, and the rest is just minimal changes to move references from the MediaWiki Skin class to the location of the new function.

https://phabricator.wikimedia.org/T257995

This PR includes:
- [ ] CI build passed

Fixes #

### DIFF
--- a/SemanticResultFormats.utils.php
+++ b/SemanticResultFormats.utils.php
@@ -44,7 +44,7 @@ final class SRFUtils {
 		];
 
 		$requireHeadItem =  [ 'srf.options' => $options ];
-		SMWOutputs::requireHeadItem( 'srf.options', SRFUtils::makeVariablesScript( $requireHeadItem, false ) );
+		SMWOutputs::requireHeadItem( 'srf.options', self::makeVariablesScript( $requireHeadItem, false ) );
 	}
 
 	/**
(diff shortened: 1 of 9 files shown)


SemanticResultFormats.utils.php
```diff
@@ -71,4 +71,19 @@ public static function htmlQueryResultLink( $link ) {
...
+	 * @param array $data
+	 *
+	 * @param string|null|bool $nonce
+	 *
+	 * @return string|WrappedString HTML
+	 */
+	public static function makeVariablesScript( $data, $nonce = null ) {
+		$script = ResourceLoader::makeConfigSetScript( $data );
+
+		return ResourceLoader::makeInlineScript( $script, $nonce );
```
JeroenDeDauw, line 86: Are these static functions both available in MW 1.31?

Ammarpad: They are, much longer before that. ResourceLoader::makeConfigSetScript() is available since MW 1.17.  The former was introduced later in 2015, (that's around MW 1.26). Also note, these static methods are what Skin::makeVariablesScript is using under the hood without any fallback since MW 1.26.

PR #640, opened 2021-02-19
Use CI scripts from SemanticMediaWiki/CIScripts

gesinn-it:
This PR aims to improve sustainability by reducing the manual effort in maintaining CI scripts and errors caused by outdated CI scripts. Instead of local Travis CI scripts it pulls centrally managed CI Scripts from SemanticMediaWiki/CIScripts. This PR can be used as a blueprint to switch all other SemanticMediaWiki repos to SemanticMediaWiki/CIScripts.

At the moment, the CI scripts are always pulled from SemanticMediaWiki/CIScripts main (master) to inject the latest updates to all consuming repos. If required, a release plan (SemVer) for SemanticMediaWiki/CIScripts can be established that allows pinning CI scripts.


JeroenDeDauw:
Why use Travis though? GH Actions probably much better going forwards

gesinn-it:
@JeroenDeDauw Github Actions might be the way to go for the future. But I think some research is required before to ensure everything is working as expected. This includes rewriting the installer scripts that currently rely on the Travis ENV.

Since most (all?) of the SemanticMediaWiki repos use Travis at the moment, switching to CIScripts would be an effort a few minutes per repo. Other contributers might be more familiar with Travis at the moment. I'd prefer doing the switch to CISctipts now to fix the other SemanticMediaWiki repos that have broken Travis at the moment.

PR #661, opened 2021-09-06
Update SRF_Calendar.php

krabina:
This PR is made in reference to: #654

I can confirm the bugfix in #654. Some more improvements might be needed, though (see mention of warning messages). However, better to have a working version that a breaking one.


kghbln:
@JeroenDeDauw :)

PR #687, opened 2021-11-24
Fix gallery rendering

### LINKED ISSUE #620: [gallery] Gallery is not neccesarily rendered immediately and needs a purge page to do so
### Setup

- MW version: 1.35
- DB (MySQL etc.): mysql
- PHP version: 7.3
- SMW version: 3.2
- SRF version: 3.2

### Issue

example:  https://www.crewsgenealogy.com/wiki/Florida_militia_muster_rolls,_Seminole_Indian_Wars.  Screen shot if it gets fixed: https://www.dropbox.com/s/7rcsg1ji35ahvaf/Screen%20Shot%202020-10-25%20at%204.59.57%20PM.png?dl=0

**Steps to reproduce**
I'm not exactly sure what causes it, but I can fix it by removing the trailing space at the end of the template on that page.   Mediawiki seems to insist on adding the space back from time to time, so I keep running into the problem.  It's not just onlthis page. 

**Expected result:**
Gallery format should be used, bot a bulleted list of images

**Observed result:**
see screenshot

gesinn-it-wam:
This PR is made in reference to: #620
Fixes #620


formats/gallery/Gallery.php
```diff
@@ -360,6 +357,20 @@ private function getImageOverlay() {
 		}
 	}
 
+	/**
+	 * Init standard widget
+	 *
+	 * @since 4.0.0
+	 *
+	 * @return array
```
JeroenDeDauw, line 365: I'd remove everything but the return. The comment suggests a side effect which is not there and the since tag is not helpful for a private function.

JeroenDeDauw: I like you extracted a function though

gesinn-it-wam: You are right, of course! I just imitated the other methods... One moment please, I have to clean up the PR!

gesinn-it-wam: Look: only the return!

JeroenDeDauw: Please choose sensible clean code over being consistent with existing code, which generally is a mess.

Btw before you invest a lot in testing result formats, you might want to have a look at https://github.com/SemanticMediaWiki/SemanticMediaWiki/pull/4313. Without that, or some similar change, it simply is not possible to write good tests.

gesinn-it-wam: I take you at your word, as this is also my preferred approach. Would you also have changed the other existing comments in this case, just to be consistent? (I would have liked to but preferred to make less noise in the end...)
Thanks for the link, I will have a look at it.

JeroenDeDauw: Can't fix everything in one go. Lots of issues. Good to focus on high prio ones. Consistency is often overrated.

PR #722, opened 2022-09-08
Update ext.srf.slideshow.js

semantisch:
The slider navigation controls wouldn't load because the module jquery.ui.slider doesn't exist anymore. Instead, check for jquery.ui, the new megamodule that replaces jquery.ui.*. See here: https://www.mediawiki.org/wiki/ResourceLoader/Migration_guide_%28users%29


krabina:
Thank you and congratulations @semantisch to your first contribution to SMW!

Issue #724, opened 2022-09-13
[pagewidget] carousel not working


krabina:
### Setup

MediaWiki | 1.35.6
-- | --
PHP | 7.4.3 (fpm-fcgi)
MariaDB | 10.3.34-MariaDB-0ubuntu0.20.04.1
ICU | 66.1
Semantic MediaWiki | 4.0.2 | GPL-2.0-or-later | Ermöglicht es, das Wiki zugänglicher zu machen – für Menschen und Maschinen 
Semantic Result Formats | 4.0.1

### Issue

Instead of the widget, a widget spinner is shown that never stops. The following errors are shown:

Wikicode used:
`{{#ask: [[Category:XYZ]] 
|format=pagewidget
|link=all
|embedformat=ul
|widget=carousel
}}`

```
The resource from “extensions/CIForms/select2/select2.full.js” was blocked due to MIME type (“text/html”) mismatch (X-Content-Type-Options: nosniff).
Loading failed for the <script> with source “extensions/CIForms/select2/select2.full.js?”.
The resource from “extensions/CIForms/select2/select2.min.css” was blocked due to MIME type (“text/html”) mismatch (X-Content-Type-Options: nosniff).
The resource from “extensions/CIForms/style.css” was blocked due to MIME type (“text/html”) mismatch (X-Content-Type-Options: nosniff).
The resource from “extensions/CIForms/validation.js” was blocked due to MIME type (“text/html”) mismatch (X-Content-Type-Options: nosniff).
```

krabina:
It has to do with chameleon skin. When I use vector, it is working.

kghbln:
Good that we at least tracked it down to SRF and Chameleon.

krabina:
According to https://github.com/ProfessionalWiki/chameleon/issues/346 this should be resolved in SRFl, not in Chameleon.

PR #727, opened 2022-09-21
SemanticResultFormats (KM-A)

thomas-topway-it:
fix issue https://github.com/SemanticMediaWiki/SemanticResultFormats/issues/724 (carousel not working)

### DIFF
--- a/formats/filtered/src/Filtered.php
+++ b/formats/filtered/src/Filtered.php
@@ -458,7 +458,9 @@ class Filtered extends ResultPrinter {
 					);
 					$viewSelectorsHtml .= Html::rawElement(
 						'div',
-						[ 'class' => "filtered-view-selector filtered-$viewName $viewid" ],
+						[ 'class' => "filtered-view-selector filtered-$viewName $viewid",
+						  'role' => "button",
+                                                'tabindex' => "0" ],
 						$viewSelectorLabel
 					);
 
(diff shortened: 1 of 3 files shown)


gesinn-it-gea:
To me this looks like the embedded jquery has been patched. This would break again on jquery update.

krabina:
What is the alternative?

gesinn-it-gea:
... well, the original source https://github.com/Wilto/Dynamic-Carousel isn't maintained anyway. So risk of breaking again is almost 0. An alternative would be to use an alternate plugin that is maintained while ideally keeping the SMW interface unchanged :-)

krabina:
Yes, I see. Well then the scope of this PR is clear: to fix the bug when bootstrap is present. Maybe we should open another issue that SRF is using an outdated plugin.

malberts:
I think this is fine (and rought what I had in mind in https://github.com/ProfessionalWiki/chameleon/issues/346#issuecomment-1249013722).

The upstream plugin is dead and if there happens to be an update there then the diff will at least show the name changed again and git blame should trace it back to this issue.

gesinn-it-gea:
... if you look at the timestamps, most of the js has not been touched the last years. As long as they work, it's fine. But fixing security issues or keeping things running with more recent versions of other components is another story. Having an additional issue to follow-up on this one is definitely the right direction.

To my opinion, this PR can be merged to get your original problem solved.

krabina:
#728

Issue #599, opened 2020-06-09
[Outline] Results with format=outline and outlineproperties displays each page only one time


Seb35:
### Setup

- MW version: 1.34.1
- DB (MySQL etc.): MySQL 5.7
- PHP version: 7.2
- SMW version: 3.1.6
- SRF version: 3.1.0
- Browsers and versions (if applicable): not applicable

### Issue

For the format "outline" with the option "outlineproperties", when a page returned by the query has multiple values for the given property, it is only displayed in the first value.

**Steps to reproduce**

For instance we have the following query:

```
{{#ask:[[Category:Article]]
|?Topic
|format=outline
|outlineproperties=Topic
}}
```

and the page "My wonderful article" contains:

```
[[Topic::My first topic]]
[[Topic::My second topic]]
[[Category:Article]]
```

**Expected result:**

The result of the query should be:

My first topic
- My wonderful article

My second topic
- My wonderful article

**Observed result:**

Currently the result of the query is:

My first topic
- My wonderful article

See [this page on sandbox.smw](https://sandbox.semantic-mediawiki.org/wiki/Query_outline_with_outlineproperties_with_multiple_values).

kghbln:
@Seb35 Thanks for reporting and working on a patch!

Seb35:
This issue occurs also when the parameter `template=` is set, and the fix seems to be similar in the file `src/Outline/TemplateBuilder.php`. I will further check this and propose another PR.

PR #600, opened 2020-06-09
Display all values with outlineproperties

Seb35:
This PR is made in reference to: #599

This PR includes:
- [x] Update of RELEASE-NOTES.md
- [x] Tests (unit/integration)
- [x] CI build passed

Fixes #599

PR #734, opened 2022-10-12
Display all values with outlineproperties (bis)

Seb35:
Similar to #600 when the parameter template= is used: the second occurence of the value was not shown.

I ran the tests locally on MW 1.35, I hope it works for other versions.

Closes: #599


Seb35:
In the tests, I copied the test SRF/Outline/Q.2 and adapted it in SRF/Outline/Q.4. I excluded in the SMW queries Q.1 to Q.3 the value [[Severity::Critical]] to avoid changing the existing results.

Issue #742, opened 2022-11-04
$node variable never set


YOUR1:
The `$node` variable in GraphPrinter
https://github.com/SemanticMediaWiki/SemanticResultFormats/blob/993be1646e4c2df3a44d91c1c26505d5b744fcb4/src/Graph/GraphPrinter.php#L191 
is not necessarily always set. Causing Fatal errors like:

```
[b21f02e9ee840c16d267cc9b] /index.php/Overzicht_RDA-klassen Error from line 190 of /var/platform/cores/1.35/20221104-1230/extensions/SemanticResultFormats/src/Graph/GraphPrinter.php: Call to a member function addField() on null

Backtrace:
#0 /var/platform/cores/1.35/20221104-1230/extensions/SemanticResultFormats/src/Graph/GraphPrinter.php(141): SRF\Graph\GraphPrinter->processResultRow(array)
#1 /var/platform/cores/1.35/20221104-1230/extensions/SemanticMediaWiki/src/Query/ResultPrinters/ResultPrinter.php(341): SRF\Graph\GraphPrinter->getResultText(SMW\Query\QueryResult, integer)
#2 /var/platform/cores/1.35/20221104-1230/extensions/SemanticMediaWiki/src/Query/ResultPrinters/ResultPrinter.php(306): SMW\Query\ResultPrinters\ResultPrinter->buildResult(SMW\Query\QueryResult)
#3 /var/platform/cores/1.35/20221104-1230/extensions/SemanticMediaWiki/includes/query/SMW_QueryProcessor.php(348): SMW\Query\ResultPrinters\ResultPrinter->getResult(SMW\Query\QueryResult, array, integer)
#4 /var/platform/cores/1.35/20221104-1230/extensions/SemanticMediaWiki/src/ParserFunctions/AskParserFunction.php(364): SMWQueryProcessor::getResultFromQuery(SMWQuery, array, integer, integer)
#5 /var/platform/cores/1.35/20221104-1230/extensions/SemanticMediaWiki/src/ParserFunctions/AskParserFunction.php(196): SMW\ParserFunctions\AskParserFunction->doFetchResultsFromFunctionParameters(array, array)
#6 /var/platform/cores/1.35/20221104-1230/extensions/SemanticMediaWiki/src/ParserFunctionFactory.php(402): SMW\ParserFunctions\AskParserFunction->parse(array)```

JeroenDeDauw:
[...]

PR #744, opened 2022-11-04
Handle unusual printout order and property chains in 'graph' format

alex-mashin:
Handle unusual printout order in 'graph' format, i.e., the one with main column absent or not the first printout.

Per @YOUR1: added 'graphfields' parameter (off by default) that switches non-page properties display to labels within nodes from edges.

Property chains are displayed as edges regardless of the type of the last link.

Unit tests for the 'graph' format rewritten and extended.

Bug: #742

### DIFF
--- a/src/Graph/GraphFormatter.php
+++ b/src/Graph/GraphFormatter.php
@@ -37,9 +37,19 @@ class GraphFormatter {
 	];
 	private $legendItem = [];
 	private $options;
+	/** @var string $lineSeparator Line separator for line wrapped long text. */
+	private $lineSeparator;
 
 	public function __construct( GraphOptions $options ) {
 		$this->options = $options;
+
+		// GraphViz is not working for version >= 1.33, so we need to use the Diagrams extension
+		// and formatting is a little different from the GraphViz extension
+		global $wgVersion;
+		$this->lineSeparator
+			= version_compare( $wgVersion, '1.33', '>=' ) && \ExtensionRegistry::getInstance()->isLoaded( 'Diagrams' )
+			? '<br />'
+			: PHP_EOL;
 	}
 
 	public function getGraph() {
@@ -102,7 +112,7 @@ class GraphFormatter {
 
 			// Display fields, if any.
 			$fields = $node->getFields();
-			if ( count( $node->getFields() ) > 0 ) {
+			if ( count( $fields ) > 0 ) {
 				$label = $nodeLabel
 					?: strtr( $this->getWordWrappedText( $node->getID(), $this->options->getWordWrapLimit() ),
 							  [ '\n' => '<br/>' ] );
@@ -250,29 +260,19 @@ class GraphFormatter {
 	 *
 	 * @return string
 	 */
-	public static function getWordWrappedText( $text, $charLimit ) {
-		$words = preg_split( '/\s+/', $text );
-		$lines = [];
-		$line = -1; // will advance to 0 on the first loop.
-		foreach( $words as $word ) {
-			if ( $line < 0 || $length + 1 + mb_strlen( $word ) > $charLimit ) {
-				// First word, or the line is getting too long. Begin a new line.
-				$lines[++$line] = []; // add new line.
-				$length = -1; // reset current line's length. No space before the first word is needed.
-			}
-			$lines[$line][] = $word;
-			$length += 1 + mb_strlen( $word ); // 1 is for the space between the words.
-		}
-		global $wgVersion;
-
-		// GraphViz is not working for version >= 1.33, so we need to use the Diagrams extension
-		// and formatting is a little different from the GraphViz extension
-		$implodeChar = version_compare( $wgVersion, '1.33', '>=' ) &&
-		\ExtensionRegistry::getInstance()->isLoaded( 'Diagrams' ) ? '<br />' : PHP_EOL;
-
-		// Glue lines by newline and words in lines by space.
-		return implode( $implodeChar, array_map( static function ( array $words ) {
-			return implode( ' ', $words );
-		}, $lines ) );
+	public function getWordWrappedText( $text, $charLimit ) {
+		return implode( $this->lineSeparator, array_reduce(
+			preg_split( '/\s+/', $text ),
+			static function ( array $wrapped, $word ) use ( $charLimit ) {
+				$last = count( $wrapped ) - 1;
+				if ( $last < 0 || mb_strlen( $wrapped[$last] ) + 1 + mb_strlen( $word ) > $charLimit ) {
+					$wrapped[] = $word;
+				} else {
+					$wrapped[$last] .= ' ' . $word;
+				}
+				return $wrapped;
+			},
+			[]
+		) );
 	}
 }
--- a/src/Graph/GraphFormatter.php
+++ b/src/Graph/GraphFormatter.php
@@ -86,7 +86,7 @@ class GraphFormatter {
 		/** @var GraphNode $node */
 		foreach ( $nodes as $node ) {
 			$instance = $this;
-			$nodeLabel = $node->getLabel();
+			$nodeLabel = htmlspecialchars( $node->getLabel() );
 
 			// take "displaytitle" as node-label if it is set
 			if ( $this->options->getNodeLabel() === GraphPrinter::NODELABEL_DISPLAYTITLE ) {
--- a/src/Graph/GraphFormatter.php
+++ b/src/Graph/GraphFormatter.php
@@ -261,18 +261,12 @@ class GraphFormatter {
 	 * @return string
 	 */
 	public function getWordWrappedText( $text, $charLimit ) {
-		return implode( $this->lineSeparator, array_reduce(
-			preg_split( '/\s+/', $text ),
-			static function ( array $wrapped, $word ) use ( $charLimit ) {
-				$last = count( $wrapped ) - 1;
-				if ( $last < 0 || mb_strlen( $wrapped[$last] ) + 1 + mb_strlen( $word ) > $charLimit ) {
-					$wrapped[] = $word;
-				} else {
-					$wrapped[$last] .= ' ' . $word;
-				}
-				return $wrapped;
-			},
-			[]
-		) );
+		preg_match_all(
+			'/\S{' . $charLimit . ',}|\S.{1,' . ( $charLimit - 1 ) . '}(?=\s+|$)/u',
+			$text,
+			$matches,
+			PREG_PATTERN_ORDER
+		);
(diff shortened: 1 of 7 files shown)


alex-mashin:
Handle unusual printout order in 'graph' format, i.e., the one with main column absent or not the first printout.

YOUR1:
What if the data of `$type` is only a `_txt`? 

Nothing will be done with the node then, even tho you want to render it.

alex-mashin:
> What if the data of `$type` is only a `_txt`?
> Nothing will be done with the node then, even tho you want to render it.
A text could be a label on a node, representing a page, but it is pointless to render a non-page property as a graph node. A "graph" consisting of disconnected texts is meaningless.

A semantic query that can be meaningfully rendered as a graph should have the `?` printout, explicit or implicit.

Moreover, previously (since the inception of this format, I guess), it was assumed that `?` was the first printout,

YOUR1:
While your merge request fixes the fatal error, it doesn't 'fix' the actual problem. 

I've got the following usecase:
Screenshot 2022-11-08 at 15 45 31

The fix suggested above doesn't show anything at all. In fact, the whole graph stays empty.

YOUR1:
Also; reverting back to the old way the `processResultRow` was done: 
```php
// loop through all row fields
foreach ( $row as $i => $resultArray ) {

	// loop through all values of a multivalue field
	while ( ( /* SMWWikiPageValue */
		$object = $resultArray->getNextDataValue() ) !== false ) {

		// create SRF\GraphNode for column 0
		if ( $i == 0 ) {
			$node = new GraphNode( $object->getShortWikiText() );
			$node->setLabel( $object->getPreferredCaption() );
			$this->nodes[] = $node;
		} else {
			$node->addParentNode( $resultArray->getPrintRequest()->getLabel(), $object->getShortWikiText() );
		}
	}
}
```

Fixed the problem for me. I'm wondering why this code was changed the way it did.

JeroenDeDauw:
This code was already complex before, and now it is even more complex. I didn't catch the bug being introduced earlier and have little confidence I will spot bugs in this code, unless I stare at it for a week. Some simplification or unit testing would help. Ideally both.

YOUR1:
> > I've got the following usecase:
> 
> What was the query?
> 
> Also, this table seems to contain no wikilinked printouts, i.e., no properties of the type 'Page'. Therefore, the graph, representing it, will have no edges. Therefore, what is the point in displaying it as a graph?

The query is 
```mediawiki
{{#ask: [[Categorie:RDA Properties]] [[Source::Agent]][[Target::Work]]
 |?Label en
 |?SubPropertyOf.Label en=(Sub) Label en
 |format=graph
 |limit=20
 |offset=0
 |link=all
 |sort=
 |order=asc
 |headers=show
 |searchlabel=... meer results
 |class=sortable wikitable smwtable
}}
```
With a older version of SRF (and GraphViz) I get:
Gebruiker_WikiSysop_digraph_QueryResult_dot

With your changes I get:
Diagrams_ab3c16d84c8671285beabf2b989b3f1f

> This code was already complex before, and now it is even more complex. I didn't catch the bug being introduced earlier and have little confidence I will spot bugs in this code, unless I stare at it for a week. Some simplification or unit testing would help. Ideally both.
I agree, the old function (as attached above) is a lot easier to understand. To be honest, I really think we need to revert the old code in to the master again, to make sure things are still workings as they have been designed before. And yes, we need a lot more unit testing if we want to implement new features.

alex-mashin:
I added the option 'graphlabel' proposed by @YOUR1 to this pull request.

I agree that this needs more unit testng.

I disagree, though, that `GraphPrinter::processResultRow()` is exceptionally complex — neither in comparison with its previous version, nor by this format's standards, nor by the standards of Semantic Result Formats or the whole SMW family.

src/Graph/GraphPrinter.php
```diff
@@ -133,79 +146,102 @@ public function getDependencyError() {
...
 
 		// GraphViz is not working for version >= 1.33, so we need to use the Diagrams extension
 		// and formatting is a little different from the GraphViz extension
+		global $wgVersion;
 		if ( version_compare( $wgVersion, '1.33', '>=' ) &&
 			\ExtensionRegistry::getInstance()->isLoaded( 'Diagrams' ) ) {
 			$result = "<graphviz>{$graphFormatter->getGraph()}</graphviz>";
 		} else {
-			// Calls graphvizParserHook function from MediaWiki GraphViz extension
+			// Calls graphvizParserHook function from MediaWiki GraphViz or External Data extension
```
YOUR1, line 170: Shouldn't this be Graphviz, Diagrams or External Data?

src/Graph/GraphPrinter.php
```diff
@@ -133,79 +146,102 @@ public function getDependencyError() {
...
 	 * Process a result row and create SRF\GraphNodes
 	 *
 	 * @since 3.1
 	 *
-	 * @param array $row
+	 * @param ResultArray[] $row
 	 *
 	 */
-	protected function processResultRow( array /* of SMWResultArray */ $row ) {
+	protected function processResultRow( array $row ) {
```
YOUR1, line 193: Your code is still giving me a fatal error, even while I'm not using graphfields=yes.

```
[d51a7ae3f8f8465ea6880001] /index.php/Gebruiker:WikiSysop Error from line 191 of /workspace/composer-ext/amSRF/src/Graph/GraphPrinter.php: Call to a member function addField() on null

Backtrace:

#0 /workspace/composer-ext/amSRF/src/Graph/GraphPrinter.php(140): SRF\Graph\GraphPrinter->processResultRow(array)
#1 /var/platform/cores/1.35/20221110-1128/extensions/SemanticMediaWiki/src/Query/ResultPrinters/ResultPrinter.php(341): SRF\Graph\GraphPrinter->getResultText(SMW\Query\QueryResult, integer)
#2 /var/platform/cores/1.35/20221110-
```

src/Graph/GraphPrinter.php
```diff
@@ -133,79 +146,102 @@ public function getDependencyError() {
...
+						'page' => $request->getLabel()
+					];
 				}
 			}
 		}
+		// Add the node, if any, its parent nodes and fields for non-Page properties to the current edge.
+		if ( $node ) {
+			foreach( $parents as $parent ) {
+				$node->addParentNode( $parent['predicate'], $parent['object'] );
+				// @TODO: add explicit nodes with hyperlinks to every parent node not added as '?', but only once.
```
YOUR1, line 238: Is this a lot of work?

src/Graph/GraphPrinter.php
```diff
@@ -133,79 +146,102 @@ public function getDependencyError() {
...
+			return $this->getDependencyError();
+		}
 
 		// iterate query result and create SRF\GraphNodes
 		while ( $row = $res->getNext() ) {
 			$this->processResultRow( $row );
 		}
 
-		// use GraphFormatter to build the graph
+		// use GraphFormater to build the graph
```
YOUR1, line 159: Typo should be; GraphFormatter

YOUR1 (changes requested):
This branch is still giving my headaches... See review.

src/Graph/GraphPrinter.php
alex-mashin: So far, I don't understand, how this happens within `if ( $node )`.

src/Graph/GraphPrinter.php
alex-mashin: Yes, and this is neither a critical error, nor relevant to this task.

src/Graph/GraphPrinter.php
alex-mashin: Corrected.

src/Graph/GraphPrinter.php
alex-mashin: Corrected the comments.

src/Graph/GraphFormatter.php
```diff
@@ -250,36 +260,19 @@ public function getGraphLegend() {
...
-			if ( $splitPosition === false ) {
-				// If there is no space (lond word), find the next space.
-				$splitPosition = strpos( $text, ' ' );
-
-				if ( $splitPosition === false ) {
-					// If there are no spaces, everything goes on one line.
-					$splitPosition = strlen( $text ) - 1;
+	public function getWordWrappedText( $text, $charLimit ) {
+		return implode( $this->lineSeparator, array_reduce(
+			preg_split( '/\s+/', $text ),
```
JeroenDeDauw, line 265: The code here is a lot simpler now, but as far as I can tell, this line is unused

src/Graph/GraphFormatter.php
```diff
@@ -250,36 +260,19 @@ public function getGraphLegend() {
...
+			static function ( array $wrapped, $word ) use ( $charLimit ) {
```
JeroenDeDauw, line 266: static function?

src/Graph/GraphFormatter.php
alex-mashin: If you mean the function `GraphFormatter::getWordWrappedText()`, it is called in at least two places and is tested.

src/Graph/GraphFormatter.php
alex-mashin: No longer static since it uses `$this->lineSeparator` set once in the constructor.

src/Graph/GraphFormatter.php
alex-mashin: And it could still be made simpler: see the new commit.

src/Graph/GraphPrinter.php
YOUR1: I can confirm its working now. I think I was on your master branch indeed, sorry for the inconveniences. I think we've created a nice solution for now. @JeroenDeDauw can you release this branch? I'll close mine.

alex-mashin:
@JeroenDeDauw?

JeroenDeDauw:
Needs rebase

alex-mashin:
> Needs rebase

Haven't you got the option to merge and rebase?

alex-mashin:
> Needs rebase

Seems rebased now.

JeroenDeDauw:
It is still not mergeable into master 

image

alex-mashin:
> It is still not mergeable into master
> 
> image

This is what I see in this pull request:
Mergeable

JeroenDeDauw:
You need to rebase with master of this repo, not with master of your fork

alex-mashin:
@JeroenDeDauw, is it mergeable now?

JeroenDeDauw:
no

alex-mashin:
I am abandoning this pull request in favour of a new one, created from scratch.

PR #745, opened 2022-11-05
Replacing unmaintained extension GraphViz with Diagrams

YOUR1:
The [GraphViz](https://www.mediawiki.org/wiki/Extension:GraphViz) extension is unmaintained and not working with MW versions newer then 1.32. [Diagrams ](https://www.mediawiki.org/wiki/Extension:Diagrams) seems to do the same, and is working properly. We've tested it and came to the conclusion that Diagrams is a good replacement.

### DIFF
--- a/formats/graphviz/README
+++ b/formats/graphviz/README
@@ -5,7 +5,7 @@ query results.
 == Installation ==
 For these two formats to work you do not just need the Semantic MediaWiki
 and the Semantic Result Formats extension to be installed but also the
-GraphViz extension and the ImageMap extension. Moreover the GraphViz extension
+Diagrams extension and the ImageMap extension. Moreover the Diagrams extension
 requires the graphviz and mscgen packages to be installed on your server.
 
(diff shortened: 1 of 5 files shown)


JeroenDeDauw:
This is a breaking change and thus requires bumping the SRF version to 5.x. This could be avoided by supporting both extensions. Not sure if that is feasable.

YOUR1:
> This is a breaking change and thus requires bumping the SRF version to 5.x. This could be avoided by supporting both extensions. Not sure if that is feasible.

I’ll see if I can come up with something 👌

YOUR1:
Don't know why the tests are failing; whenever I run the test locally, it succeeds. I'll see what's up tomorrow.

gesinn-it-gea:
To my opinion, it would be okay to release SRF 5.x and end the tradition, that SRF always has the same major verison as SMW.

JeroenDeDauw:
Totally agree, we have stopped following that tradition a while ago and are only coincidentally on the same major version now.

PR #746, opened 2022-11-09
GraphViz bug fixes

YOUR1:
- Fixed a bug in the `getWordWrappedText()` function where `<br />` would be shown instead of a `\n`
- Added documentation fields in GraphOptions and fixed a bug where invalid characters would break the generation of graphs. 
- Also added a unit test for that same getGraphName function.


YOUR1:
Sorry for the overload on commits.. Guess I'm to tired :)

YOUR1:
@JeroenDeDauw can you review/merge?

PR #747, opened 2022-11-10
GraphViz Added graphfields option

YOUR1:
Reverting back changes from https://github.com/SemanticMediaWiki/SemanticResultFormats/commit/dcfd97fc12b153a6ea0d227acde1475079bc1990 to the original one since it breaks the functionality that was here in the first place.

Since I don't want to neglect the changes @alex-mashin made, I implemented a 'graphfields' option, so his fields feature are still visible and usable.

Graph without fields (old and default behaviour):
Screenshot 2022-11-10 at 14 58 48

Graph with fields (new option):
Screenshot 2022-11-10 at 15 00 16

### DIFF
--- a/src/Graph/GraphFormatter.php
+++ b/src/Graph/GraphFormatter.php
@@ -100,7 +100,7 @@ class GraphFormatter {
 
 			// Display fields, if any.
 			$fields = $node->getFields();
-			if ( count( $node->getFields() ) > 0 ) {
+			if ( count( $fields ) > 0 && $this->options->showGraphFields() ) {
 				$label = $nodeLabel
 					?: strtr( $this->getWordWrappedText( $node->getID(), $this->options->getWordWrapLimit() ),
 							  [ '\n' => '<br/>' ] );
--- a/src/Graph/GraphOptions.php
+++ b/src/Graph/GraphOptions.php
@@ -28,6 +28,9 @@ class GraphOptions {
 	private $showGraphColor;
 	private $showGraphLegend;
 
+	/** @var bool */
+	private $showGraphFields;
+
 	public function __construct( $options ) {
 		$this->graphName = trim( $options['graphname'] );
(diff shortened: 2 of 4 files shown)


alex-mashin:
How will your code process the query below?
```
{{#ask: [[Categorie:RDA Properties]] [[Source::Agent]][[Target::Work]]
 |?Label en
 |?SubPropertyOf.Label en=(Sub) Label en
 |?
 |format=graph
}}
```

alex-mashin:
And no, the graph with fields is now meaningless. In the query above with `graphfields`, in the edges should represent 'SubpropertyOf' (if it is of the type 'Page'), and 'Label en' and 'Description en' (the latter is missing altogether) should only be fields within the nodes.

YOUR1:
> How will your code process the query below?
> 
> ```
> {{#ask: [[Categorie:RDA Properties]] [[Source::Agent]][[Target::Work]]
>  |?Label en
>  |?SubPropertyOf.Label en=(Sub) Label en
>  |?
>  |format=graph
> }}
> ```

Diagrams_c8a8e58b95c9638a1bd81709c99b3882

> The legend for graphs reads 'black: label en'. It should be 'Black: SubpropertyOf'.
> 
> And 'Label en' has only one value per property, as the table shows, yet there are two and three at the nodes on the second graph.

I havent changed anything about the code regarding the legend, thus I think that's a bug that was already there.

--
To keep things aligned with the merge request - I think we need to revert to this point so the code that was here originally works again. I suggest that the bugs in the legend, or additional changes in the fields mechanism can be fixed after the code has been merged to master.

YOUR1:
@JeroenDeDauw can you confirm this is the right strategy?

YOUR1:
We've fixed the issue I've had on https://github.com/SemanticMediaWiki/SemanticResultFormats/pull/744. This can be closed. Thanks @alex-mashin.

PR #749, opened 2022-11-20
Handle unusual printout order and property chains in 'graph' format

alex-mashin:
Handle unusual printout order in 'graph' format, i.e., the one with main column absent or not the first printout.

Per @YOUR1: added 'graphfields' parameter (off by default) that switches non-page properties display to labels within nodes from edges.

Property chains are displayed as edges regardless of the type of the last link.

Unit test for the 'graph' format rewritten and extended.

Bug: #742

### DIFF
--- a/tests/phpunit/Unit/Graph/GraphFormatterTest.php
+++ b/tests/phpunit/Unit/Graph/GraphFormatterTest.php
@@ -16,78 +16,26 @@ use SRF\Graph\GraphOptions;
  */
 class GraphFormatterTest extends \PHPUnit\Framework\TestCase {
 
-	/*
-	* @see https://www.semantic-mediawiki.org/wiki/Help:Graph_format
-	*/
-	private $options;
-
-	private $graphFormatter;
-
-	private $nodes = [];
-
-	protected function setUp(): void {
-		parent::setUp();
-
-		$params = [
-			'graphname' => 'Unit Test',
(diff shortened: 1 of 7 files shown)


alex-mashin:
This comment is to bring this pull request to the attention of @JeroenDeDauw and @YOUR1.

YOUR1:
Sorry for the hold up. I've tested it and found **no** further issues using the default (without |graphfields) graph. Same goes for the new functionality; graphfields - no issues there aswel.

tests/phpunit/Unit/Graph/GraphFormatterTest.php
```diff
@@ -16,89 +16,225 @@
...
+			$graph_node = new GraphNode( $node['name'] );
+			$graph_node->setLabel( $node['label'] );
+			if ( isset( $node['parents'] ) ) {
+				foreach ( $node['parents'] as $parent ) {
+					$graph_node->addParentNode( $parent['predicate'], $parent['object'] );
+				}
+			}
+			if ( isset( $node['fields'] ) ) {
+				foreach ( $node['fields'] as $field ) {
+					$graph_node->addField( $field['name'], $field['value'], $field['type'], $field['page'] );
```
JeroenDeDauw, line 133: Control structures in tests are a sign of something going wrong. Been a while since I saw a case with trippy nested control structures.

Issue #753, opened 2022-12-21
Translation issue with message MediaWiki:Srf-paramdesc-carousel-slick-option/qqq


verdy-p:
URL: https://translatewiki.net/w/i.php?title=Special:Translate&showMessage=srf-paramdesc-carousel-slick-option&group=mwgithub-semanticresultformats&language=qqq

See the FIXME added in the documentation of this message. For now the translation is not clear, and the doc is insufficient. This is a new message, so it may still be in beta (just like the carousel visualization widget which has demos in the SMW wiki, but none of them using any "slick option", and none of them for the "table" output format but only for galeries).

These carousel widgets also have other unsolved maintenance issues listed in GitHub (if changes are made, this could alter the behavior or layout, and would require change in the documentation of this feature).

For now it's not clear how to translate "slick", but I suppose this means that it controls whever the carrousel slides smoothly from page to page, or instantly. All existing demos for now seem to use a smooth ("slick"?) sliding behavior by default. But "slick" also has another meaning in English (as a name), and may refer to a layout rather than a sliding behavior.

Other possible layout or behavior (for galleries) would use a freely sliding area (working like a scrollbar, where the content could still be loaded on demand, when it becomes visible, or fully preloaded), instead of using "paged" results. And MediaWiki also featuers its own galleries, that are "playable" in full screen, paged image by image, to show all images used in any page. In-content galleries also feature various layouts (using vertical scrolling if needed, but not requiring any javascript addon). Getting galleries with enhanced layouts or behavior is still a work ni progress, not just in SMW but in Mediawiki in general. It is also sensitive in terms of usability/accessibility, and in terms of cross-browser and cross-device compatibility. Other sites (notably search engines, photo-based social networks, dating sites & apps, but also merchant sites for listing their products) also propose their own layouts and behaviors for showing and navigating across large image galleries.

kghbln:
@krabina @thomas-topway-it This is about pull https://github.com/SemanticMediaWiki/SemanticResultFormats/pull/730

If you just explain in a short sentence what is slick about the slick option.... I could update the system message docu on twn.

thomas-topway-it:
Slick is the Javascript library on which the carousel format is based

https://github.com/kenwheeler/slick

all the slick's options are supported using the prefix slick- in parameters, for instance slick-autoplaySpeed

verdy-p:
[...]

PR #730, opened 2022-09-29
 SemanticResultFormats datatables upgrade (KM-A)

thomas-topway-it:
This PR is made in reference to: [KM-A Knowledge Management Associates](https://km-a.net/knowledge-wiki/) upgrade of SemanticResultFormats Datatables
This PR addresses or contains:

- upgrade of Datatables to 1.12.1, Buttons 2.2.3, Select 1.4.0
- fixes https://github.com/SemanticMediaWiki/SemanticResultFormats/issues/188
- fixes https://github.com/SemanticMediaWiki/SemanticResultFormats/issues/726
- fixes https://github.com/SemanticMediaWiki/SemanticResultFormats/issues/721 
(added parameters: pagelength and lengthmenu)
- fixes https://github.com/SemanticMediaWiki/SemanticResultFormats/issues/93
- (datatables) added support for basic theme 
- adds play/pause button on slideshow navigation
- adds a new result format (carousel based on [slick](https://kenwheeler.github.io/slick/))
- updates Javascript tests for Datatables format, and integration tests for Carousel format
- fixes https://github.com/SemanticMediaWiki/SemanticResultFormats/issues/750

PR #754, opened 2022-12-22
Improve system messages

kghbln:
A follow-up to #730 

Fixes: #753


i18n/en.json
```diff
@@ -108,9 +108,9 @@
 	"srf-paramdesc-renderer": "Select a graph/chart renderer",
 	"srf-paramdesc-filling": "Individual filling option",
 	"srf-paramdesc-theme": "Select a grid theme",
-	"srf-paramdesc-pagelength": "define datatable page length",
-	"srf-paramdesc-lengthmenu": "define datatable length menu",
-	"srf-paramdesc-datatables-columnstype": "define datatable columns type",
+	"srf-paramdesc-pagelength": "Specifies the length of the page for the datatable",
+	"srf-paramdesc-lengthmenu": "Specifies the length of the menu for the datatable",
+	"srf-paramdesc-datatables-columnstype": "Specifies the type for the columns of the datatable",
```
JeroenDeDauw, line 113: Isn't the "for the datatable" part already clear from context? And should it not be "of the datatable" in all cases?

`length of the page` can be shortened to `page length`

kghbln: Sure, why not.

PR #756, opened 2022-12-23
Fix typo

### LINKED ISSUE #755: Translation issue with message MediaWiki:Srf-paramdesc-carousel-width/en
URL: https://translatewiki.net/w/i.php?title=Special:Translate&showMessage=srf-paramdesc-carousel-width&group=mwgithub-semanticresultformats

Typo: replace "with" by "width" in
> `Specifies the with of the carousel`

kghbln:
fixes #755

Thanks @verdy-p for discovering it.

### DIFF
--- a/i18n/en.json
+++ b/i18n/en.json
@@ -285,7 +285,7 @@
 	"srf-paramdesc-zoom": "Enable zoom",
 	"srf-paramdesc-carousel-class": "Specifies the CSS class for the carousel",
 	"srf-paramdesc-carousel-width": "Specifies the width of the carousel",
-	"srf-paramdesc-carousel-height": "Specifies the hight of the carousel",
+	"srf-paramdesc-carousel-height": "Specifies the height of the carousel",
 	"srf-paramdesc-carousel-captionproperty": "Specifies the name of a semantic property present on the queried pages to be used as caption",
 	"srf-paramdesc-carousel-titleproperty": "Specifies the name of a semantic property present on the queried pages to be used as title",
 	"srf-paramdesc-carousel-linkproperty": "Specifies the name of a semantic property present on the queried pages to be used for links",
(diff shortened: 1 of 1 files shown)


verdy-p:
In your recent commit I see a similar typo, one line below in your JSON file
> "hight" instead of "height"
I did not remark it in translatewiki.net (because someone else translated it)

kghbln:
Aaargh, I always mess these two up. Web editing via laptop does not help either. :|

verdy-p:
That's jsut like the reply I amde this morning on my Android phone: there are too many typos introduced repeatedly by the Android message editor, mixing languages and "autofixing" English, even after I rejected their automatic changes. Sometimes the message is read multiple times, but these fixes are made as soon as we loose the focus on the text input, just as we press the "Send" button (in fact case the Android email app does not let us confirm these changes, the email is sent anyway).

There are of course many words that we frequently type with some inverted letters (notaby when typing fast with fingers of the two hands on both sides of the keyboard). Here also we do not always see that because we were focused on typing the rest of the sentence. Rereading is often better when it is performed a bit later, and when the text is rendered using a different presentation. An initial "preview" mode often helps to prevent this, but not always because the presentation is also too similar: we sometimes need a different context of use.

For that reason also, translations made in TWN also need independant reviewers (any one can make such typos, even when they try to be very careful and all seems OK on the first views made in the same short period of time and on the same screen). But I also regularly reread my own typed messages, on different days when I need to refocus on the topic with a fresh mind not using the many "shortcuts" we use constantly without thinking about them: typing something takes other intellectual resources not needed when we are only reading (and less distracted by simultaneous actions or thoughts): copy-editing can be done by ourself if it's not done by someone else, but cannot be done reliably by ourselves on the same short period. We just hope that we do things correctly the first time.

PR #759, opened 2023-01-29
[eventcalendar] Improve styleability of event tooltip

gesinn-it-gea:


gesinn-it-gea:
This is how the popup looks after the change:

grafik

PR #771, opened 2023-03-16
Datatable v2 (rebased)

thomas-topway-it:
A complete refactoring of the datatables result format implementing a dedicated api and a SMW <code>BeforeQueryResultLookupComplete</code> hook in order to perform some preprocessing of the table data.

It includes the legacy datatables format with the format name <code>datatables-legacy</code>

Here is the complete documentation

https://www.semantic-mediawiki.org/wiki/Help:Datatables_format_(v2)

Key features:

- cells content fully consistent with the standard SMW table format (including the use of [Printouts format](https://www.semantic-mediawiki.org/wiki/Help:Displaying_information))
- can access all the rows of a query using Ajax when necessary
- supports navigation, filtering, and sorting against the complete data-set (using Ajax when necessary)
- supports [sorting](https://www.semantic-mediawiki.org/wiki/Help:Sorting) using multiple columns
- a wide range of native Datatables options are accessible through the askquery


kghbln:
@thomas-topway-it Does this replace the existing datatabel format, or is this an additional and separate implementation?

thomas-topway-it:
Hello Karsten, I was just trying to rename https://www.semantic-mediawiki.org/wiki/Help:Datatables_format_(v2) to https://www.semantic-mediawiki.org/wiki/Help:Datatables_format, after moving https://www.semantic-mediawiki.org/wiki/Help:Datatables_format to https://www.semantic-mediawiki.org/wiki/Help:Datatables_format_(legacy)

but perhaps the redirect has to be deleted before ?

So it replaces the old datatable format but at the same time the old datatable format is accessible using `format=datatables-legacy`

However, it is not a breaking change, you can just purge the page without to edit the askquery and it will work with the new version !
Let me know if there is any issue !

kghbln:
Thanks for the info. The docu issues have more or less been sorted out.

To cut it short: It is an additional result printer, and to me, it is indeed breaking for users: All current users are using `format=datatables` all over the place, which will now show the new printer instead of the legacy printer. I do not know how the new standard datatables compare to datatables-legacy when using the drop-in replacement. Still, we should inform people of this in the wiki docu and via the release notes once SRF 4.1.0 gets published.

thomas-topway-it:
hello Karsten, so I think the replacement happened in a natural way because at a first stage of development we did not want to replace the previous datatables format and to create a new format, but to enhance the previous one. Following, it required a substantial rewriting that led to the removal of some of the existing features, like the bootstrap theme, the "query-panel" and "instant update". For this reason I've preferred to keep the old format using the name "datatables-legacy". The drop-in replacement should be seamless apart for these features, and the bugs affecting the display of data should have been fixed. Of course I can contribute to the release notes when SRF 4.1.0 will be published, and if you note something that doesn't work properly with regard of the "drop-in replacement" itself we can try to take the required measures promptly.

thomas-topway-it:
hello Karsten, ok great, Actually we could have named the new datatables format "datatables-v2" and to leave the previous as it is, but the current solution seems more natural even if it implies some change. Also, the work was directed by @krabina so it complies with his requirements.

kghbln:
Hi Thomas. it is fine as it is unless we inform users of the software about it. I have done so on the smw wiki and once SRF 4.1 we can also do so in the release notes.. All good. :)

PR #774, opened 2023-04-28
Datatables v2 SearchPanes client/server side

thomas-topway-it:
- Fixes https://www.semantic-mediawiki.org/wiki/Help:Datatables_format#Limitations (searchPanes now works in conjunction with Ajax, i.e. with distinct values for each pane computed server-side based on the actual query)
- Fixes https://www.semantic-mediawiki.org/wiki/Help:Datatables_format#Known_issues (SearchPanes work correctly also if cells contain html)

- (redesigned) the parameter `datatables-searchPanes ` is now used to toggle on/off searchPanes independently by the "P" symbol in `datatables-dom`
- adds custom parameter `datatables-searchPanes.minCount` (to include in searchPanes only items with occurrence greater or equal than `minCount`) 
- adds custom parameter `datatables-searchPanes.htmlLabels` to toggle html/text only in panes' entries
- updated default separator for multiple values in cells (now is `,&#32;`) 
- searchPanes support category printout
- searchPanes support dates ranges, and other grouping performed using [printout formats](https://www.semantic-mediawiki.org/wiki/Help:Displaying_information)

demos:
* categories and dates range: https://fina.knowledge.wiki/Test_Datatables/SearchPanes_Categories
* `Noajax` (all data are retrieved on page load) https://fina.knowledge.wiki/Catalogues_of_Numismatic_Collections
* multiple filters https://fina.knowledge.wiki/Test_Datatables


JeroenDeDauw:
Some UX feedback: when selecting a filter on the left, I would find it more intuitive for the filters on the right to also update.

image

Issue #764, opened 2023-02-06
[eventcalendar] fails when properties have a preferred label


krabina:
### Setup

- MW version: 1.35.8
- PHP version: 7.4
- SMW version: 4.0.1
- SRF version: 4.0.1

### Issue
A working eventcalendar fails to display correctly (only shows the header) as soon as a property is present in the printout that has a preferred property label set. If you remove the "has preferred property label" statement, the exact same query is working again.

**Steps to reproduce**

1. Create a working example for an eventcalendar
2. Add some printout property
3. add _has preferred property label_ statement to a property used in the query

**Expected result:** ...
eventcalendar should show and display the property in the defined label

**Observed result:** ...
eventcalender only showsl the header, without and days or weeks in the month or week view, only displays something when you switch to "day"

Following message in the browser console:

```
Uncaught TypeError: list[property] is undefined
    getTypeId extensions/SemanticResultFormats/resources/ext.srf.api.results.js?ba572:138
    normalize extensions/SemanticResultFormats/resources/ext.srf.api.query.js?d13fd:169
    each jQuery
    normalize extensions/SemanticResultFormats/resources/ext.srf.api.query.js?d13fd:163
    type extensions/SemanticResultFormats/resources/ext.srf.api.query.js?d13fd:188
    init extensions/SemanticResultFormats/formats/calendar/resources/ext.srf.formats.eventcalendar.js?d3a86:787
    <anonymous> extensions/SemanticResultFormats/formats/calendar/resources/ext.srf.formats.eventcalendar.js?d3a86:1102
    jQuery 4
        fire
        fireWith
        tuple[0]
        using
    enqueue load.php?debug=true&lang=de&modules=startup&only=scripts&raw=1&skin=chameleon:1284
    using jQuery
    <anonymous> extensions/SemanticResultFormats/formats/calendar/resources/ext.srf.formats.eventcalendar.js?d3a86:1101
    jQuery 2
        each
        each
    <anonymous> extensions/SemanticResultFormats/formats/calendar/resources/ext.srf.formats.eventcalendar.js?d3a86:1047
    jQuery 2
        mightThrow
        process
ext.srf.api.results.js:138:6
    getTypeId extensions/SemanticResultFormats/resources/ext.srf.api.results.js?ba572:138
    normalize extensions/SemanticResultFormats/resources/ext.srf.api.query.js?d13fd:169
    each jQuery
[...]

PR #777, opened 2023-05-30
preferred label and eventcalendar

thomas-topway-it:
fixes https://github.com/SemanticMediaWiki/SemanticResultFormats/issues/764

***attention! the proposed solution is a workaround for the use with SRF's `ext.srf.api.query.js -> toList` (a regex to match the printout label and custom label), however a preferred solution would be at SMW level trying to fix PrintRequest's `Serializer -> doSerializeProp`

### DIFF
--- a/src/ResourceFormatter.php
+++ b/src/ResourceFormatter.php
@@ -5,6 +5,7 @@ namespace SRF;
 use Html;
 use SMWOutputs as ResourceManager;
 use SMWQueryResult as QueryResult;
+use SMW\Query\PrintRequest;
 
 /**
  * @since 3.0
@@ -79,7 +80,7 @@ class ResourceFormatter {
 	}
 
 	/**
-	 * @param \SMW\Query\PrintRequest[] $printRequests
+	 * @param PrintRequest[] $printRequests
 	 * @param array $ask
 	 * @return array
 	 */
--- a/src/ResourceFormatter.php
+++ b/src/ResourceFormatter.php
@@ -78,6 +78,46 @@ class ResourceFormatter {
 	}
 
 	/**
+	 * @param array $printRequests
+	 * @param array $ask
+	 * @return array
+	 */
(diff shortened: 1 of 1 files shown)


src/ResourceFormatter.php
```diff
@@ -77,6 +78,51 @@ public static function encode( $id, $data ) {
 		);
 	}
 
+	/**
+	 * @param array $printRequests
+	 * @param array $ask
+	 * @return array
+	 */
+	private static function appendPreferredPropertyLabel( $printRequests, $ask ) {
```
JeroenDeDauw, line 86: Better to use types in PHP and remove the doc block

JeroenDeDauw: Tho you could keep the doc block and add more detailed type info such as `PrintRequest[]` or whatever the class name actually is.

thomas-topway-it: I've added the param type
` * @param PrintRequest[] $printRequests`

JeroenDeDauw: Without it being imported, this references to SRF\PrintRequest, which does not exist. 

You need to make sure the type is actually correct. Otherwise it does not help by providing autocompletion and error detection in the IDE. PhpStorm can autocomplete on the type in the comment and even import it automatically for you.

thomas-topway-it: so it should rather be @param `SMW\Query\PrintRequest[] $printRequests` ?

JeroenDeDauw: That would resolve to `SRF\SMW\Query\PrintRequest[]`. You need a backslash at the front: `\SMW\Query\PrintRequest[]`

Though better is to import the name with an import statement, so you do not have the FQN in the docs.

thomas-topway-it: I have added the import statement

PR #787, opened 2023-08-15
Remove SMW requirement from composer.json

yaronkoren:
Of course SMW is required for SRF to run, but the presence of this line makes it impossible for SMW and SRF to both be downloaded via a non-Composer option like Git, because then calling "composer install" for SRF, in order to get its dependencies, will lead to SMW getting re-downloaded. The line doesn't seem necessary, anyway.


hexmode:
If the reason for this is because of a problem with composer vs git, why not just use `--prefer-source` when you run composer?  That would get the SMW dependency via git.

yaronkoren:
That sounds better, but I think better yet would be to not get SMW at all. Why get it?

hexmode:
> That sounds better, but I think better yet would be to not get SMW at all. Why get it?

It is a dependency and composer is a dependency manager, so naturally, it gets it.

If this PR is accepted, other people will come along and say "Oh, the SMW dependency is missing.  Let me add it."  People will be continually pointed back here to say "We're not including it so that composer DTRT in this one case.  Yes, we know that it is wrong in some other cases, but we're not going to fix it.  You just always have to specify SMW *and* SRF."

yaronkoren:
Well, disregarding for now any difficulties with future maintenance, what would you say is the preferred behavior? I think there's a strong case that Composer "require" should only be used for libraries (including "extension libraries" like DataValues), and not for real MediaWiki extensions, because those are unnecessary and can lead to unexpected behavior.

hexmode:
> ... and not for real MediaWiki extensions, because those are unnecessary and can lead to unexpected behavior.

I'm not sure what you mean here by "unnecessary". SMW is a necessary dependency for SRF, isn't it?

yaronkoren:
I mean that it's not necessary because the user can always include a line for these extensions in their composer.json file in they want them to be downloaded during this operation. (And presumably, in the vast majority of cases, they already will.) A user who _doesn't_ want an extension like SMW to be downloaded during the operation (because they've already downloaded it separately), on the other hand, will have no recourse, as far as I can tell.

krabina:
Your thoughts, @gesinn-it-gea?

gesinn-it-gea:
... I follow @yaronkoren reasoning. In my experience, dependency managers are good for managing libraries, but don't necessarily make things easier for installing extensions. Personally, I prefer to be able to take care of the dependencies myself.

yaronkoren:
I've now restored this patch, and, per https://github.com/SemanticMediaWiki/SemanticMediaWiki/issues/5892, I'm hoping that it will now be accepted.

PR #788, opened 2023-09-04
Harmonize resource loading spinner

thomas-topway-it:
SemanticResultFormats were using 2 different spinners, `srf-loading-dots`, used for instance in datatables, and a more convoluted (or unattractive) spinner `srf-spinner` as in the picture.

semantic-result-formats-alternate-spinner

the present commits unifies the 2 different spinners using always the following

semantic-result-formats-spinner
The old spinner tried to take into account the container size, but fails to do so in many cases since the container size isn't fixed in all cases and is determined by the rendered widget, so where this feature is envisaged this commit could be followed by a further enhancement, to ensure that all containers have fixed width, and to use the standard `mw-rcfilters-spinner` (see picture below) in these cases. (or as a replacement of the spinner now used).

semantic-result-formats-spinner-new


thomas-topway-it:
@gesinn-it-gea

Issue #784, opened 2023-07-27
[carousel] slick-slidesToShow does not show up multiple images


michaelfreiberg:
### Setup

- MW version: 1.35.11
- DB (MySQL etc.): 10.6.14-MariaDB
- PHP version: 7.4.33 (apache2handler)
- SMW version: 4.1.1
- SRF version: 4.0.2

### Issue
Setting the parameter `slick-slidesToShow` in the result format _carousel_ seems to have no effect. I used the following query:
```
{{#ask: [[DIGISAM-ID::+]]
|?DIGISAM-Titelbild
|?DIGISAM-Bildunterschrift
|?Erscheinungsjahr
|format=carousel
|limit=999
|sort=Erscheinungsjahr
|order=asc
|imageproperty=DIGISAM-Titelbild
|titleproperty=DIGISAM-Bildunterschrift
|width=200px
|slick-slidesToShow=4
|slick-slidesToScroll=1
|slick-autoplaySpeed=1500
}}
```

**Expected result:**
Setting the parameter `slick-slidesToShow` to a value higher than 1 should show multiple images in the slick carousel.

My current workaround is to hard-code the default value of `slick-slidesToShow` to "4" in resources/slick.js and to fix the width of the `slick-slider` class to "800px" with an import CSS rule in MediaWiki:Common.css. The working result can be seen [here](https://www.fruehe-texte-holocaustliteratur.de). My guess is that the width parameter of the #ask query sets **both** the width of the image (CSS class `slick-slide-content`) and the slider (CSS class `slick-slider`) but it should set the width of the slider to a multiple of the image width (image width * slidesToShow).

thomas-topway-it:
@michaelfreiberg I have sent a pull request that should solve the problem, although it may require more testing https://github.com/SemanticMediaWiki/SemanticResultFormats/pull/793

(it also implements your suggestion: note however that the width parameter is considered the width of the entire carousel, not of the single image.)

See also here https://github.com/kenwheeler/slick/issues/1071
and you might also play with the `slick-variableWidth` parameter

PR #793, opened 2023-10-31
Fix issue 784 slidesToShow

thomas-topway-it:
- fixes https://github.com/SemanticMediaWiki/SemanticResultFormats/issues/784
- fixes default arrow color

the issue was caused by 2 concurrent problems, the parameter `slick-responsive` with wrong default value prevented all subsequent parameters to work, including `slick-slidesToShow`, now all the parameters listed here https://www.semantic-mediawiki.org/wiki/Help:Carousel_format should work correctly.
It has also been implemented the suggestion of @michaelfreiberg mentioned in the same issue

### DIFF
--- a/formats/carousel/Carousel.php
+++ b/formats/carousel/Carousel.php
@@ -4,7 +4,7 @@
  *
  * @license GPL-2.0-or-later
  *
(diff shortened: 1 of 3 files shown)


krabina:
@gesinn-it-gea is this ready to merge?

PR #798, opened 2023-11-28
Update ci.yml

kizule:
Removed MediaWiki 1.40 because of:

1. I don't remember that something related to Semantic MediaWiki has ever officially supported a non-LTS version of MediaWiki
2. Tests are failing for MediaWiki 1.40, it's not worth fixing it as MediaWiki 1.40 is non-LTS, because of previous reasons.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -38,13 +38,6 @@ jobs:
             database_image: "mariadb:latest"
             coverage: false
             experimental: true
-          - mediawiki_version: '1.40'
-            smw_version: dev-master
-            php_version: 8.1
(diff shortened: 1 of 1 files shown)


gesinn-it-gea:
The question is: should we include the latest non-LTS version to indicate that there may be some open issues with a particular version? In my opinion, testing should reveal potential problems before the next LTS version is released. Unfortunately, the official MediaWiki Docker images only appear with the release, so CI could only be adapted to them very late.

What exactly bothers you about the failing CI for 1.40. The red cross?

grafik

gesinn-it-gea:
I have removed 1.40 now

PR #803, opened 2023-12-07
Fix GraphViz extension support

JeroenDeDauw:


krabina:
Ah, I see. So Extension:Diagrams does work instead of Extension:GraphViz?

Issue #818, opened 2024-02-27
bug (filtered): use a more specific term for the CSS classes "enabled", "disabled" 


gesinn-it-gea:
By default, each filter has assigned the css class `enabled`. This is related to the filter option `+value filter switches=on off`.

Example:
`<div id="hn85zw4n6q" class="filtered-filter filtered-value enabled">`

"enabled" is a very general term that runs the risk of also being used elsewhere. Specifically, "enabled" is also used by leaflet maps, for example.

Replace "enabled" with "filtered-filter-enabled" and "disabled" with "filtered-filter-disabled"

gesinn-it-gea:
![image](https://github.com/SemanticMediaWiki/SemanticResultFormats/assets/10398316/f48acc3c-082b-4d24-8a54-ef1a61524d0c)

gesinn-it-gea:
The `enabled` class is related to the `+value filter switches=on off` switch and is `on` by default.

PR #837, opened 2024-06-10
add more specific term for CSS classes 'enabled' and 'disabled'

gesinn-it-ilm:
Terms for CSS classes '**enabled**' and '**disabled**' has been updated and changed to more specific terms. 

- 'enabled'  => 'filtered-filter-enabled';
- 'disable' => 'filtered-filter-disable'


gesinn-it-gea:
closes #818

Issue #778, opened 2023-06-17
Replace removed core MW function wfFindFile()


D-Groenewegen:
### Setup

- MW version: 1.39

### Issue

An issue was reported to the mailing list by Joseph Padfield, who observed he could no longer use the Dygraphs format with `datasource=file`. I was able to confirm the issue and it appears that the result format still relies on wfFindFile(), which was deprecated in MW 1.34 and finally removed in MW 1.39. 

### Solution
The following files require an update :
- https://github.com/SemanticMediaWiki/SemanticResultFormats/blob/master/formats/dygraphs/SRF_Dygraphs.php
- https://github.com/SemanticMediaWiki/SemanticResultFormats/blob/master/formats/media/MediaPlayer.php#L370

See https://www.mediawiki.org/wiki/Manual:GlobalFunctions.php#wfFindFile()
` Use MediaWikiServices::getInstance()->getRepoGroup()->findFile() instead`

D-Groenewegen:
Should be fixed with the commit above.

PR #838, opened 2024-06-12
Updates to wfFindFile again

D-Groenewegen:
No need to support the older method wfFindFile (pre-MW 1.34). Since SRF 4.0.x, the minimum version is MW 1.35. Also corrected an earlier oversight.
#778


alexander-gesinn:
Thanks for your PR! We need to explain the CI and testing a little bit better :-) We now have Docker Compose based CI that can be used for testing locally by simply running "make ci". See https://github.com/gesinn-it-pub/docker-compose-ci

PR #841, opened 2024-07-15
Update README.md

kghbln:
- update packgist badges (stable version, download count)
- typos and tweaks


kghbln:
@JeroenDeDauw Unsure about the "but can also be executed locally using the shortcut command `composer phpunit` from the extension base directory." part in the testing section. This could be dropped.

PR #845, opened 2024-07-22
WCAG improvement for format=datatable

robinwiel:
We would like to include the "aria-live" attribute (set to "polite") on the tbody-element of a DataTable.

As the content inside the tbody-element can dynamically change (e.g. by changing the page of the results, or by entering a keyword in the search bar), the "aria-live" attribute will inform screen readers that the content of the tbody-element has changed.

See https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Live_Regions


YOUR1:
Nice!

@gesinn-it-ilm can you approve? ;-)

Issue #853, opened 2024-10-17
Error when using intro or outro parameter


YvarRavy:
### Setup

- MW version: 1.39
- DB (MySQL etc.): i don't think that matters here
- PHP version: 7.4
- SMW version: 4.1.3
- SRF version: 4.2.1

### Issue
When trying to add either an intro or outro text and search for the results with format datatables you get a big error

**Steps to reproduce** (recommendation is to use the [sandbox](https://sandbox.semantic-mediawiki.org))
the sandbox gives a different error because of the Extension:PageInCat
but the way to reproduce the error on any other wiki with SMW & SRF
```
1. Create ask query doesn't really mattter what i just use [[Category:+]]
2. Use Format=Datatables
3. under Parameters fill in either the intro or outro input field
5. fetch results
```

**Expected result:** 
the normal results with the outro and intro added to the front and back
**Observed result:**
Fatal error
**Stacktrace** 
```
[c127aab4f08c39dcf86f038c] /index.php?title=Speciaal:Vragen& Error: Call to a member function getMaxIncludeSize() on null

Backtrace:

from mediawiki/includes/parser/Parser.php(2937)
#0 mediawiki/includes/parser/Parser.php(1609): Parser->replaceVariables(string)
#1 mediawiki/includes/parser/Parser.php(881): Parser->internalParse(string, boolean, boolean)
#2 mediawiki/includes/parser/Parser.php(905): Parser->recursiveTagParse(string, boolean)
#3 composer-ext/SemanticResultFormats/formats/datatables/DataTables.php(483): Parser->recursiveTagParseFully(string)
#4 composer-ext/SemanticResultFormats/formats/datatables/DataTables.php(445): SRF\DataTables->handleNonFileResult(string, SMW\Query\QueryResult, integer)
#5 composer-ext/SemanticMediaWiki/src/Query/ResultPrinters/ResultPrinter.php(304): SRF\DataTables->buildResult(SMW\Query\QueryResult)
#6 composer-ext/SemanticMediaWiki/src/MediaWiki/Specials/SpecialAsk.php(500): SMW\Query\ResultPrinters\ResultPrinter->getResult(SMW\Query\QueryResult, array, integer)
[...]

PR #854, opened 2024-10-17
fix error when using intro and outro text

YvarRavy:
Fixes issue: https://github.com/SemanticMediaWiki/SemanticResultFormats/issues/853
i first checked how SMW handels this with broadtable and pretty much just moved this code

### DIFF
--- a/formats/datatables/DataTables.php
+++ b/formats/datatables/DataTables.php
@@ -475,12 +475,20 @@ class DataTables extends ResultPrinter {
 
 		// Apply intro parameter
 		if ( ( $this->mIntro ) && ( $results->getCount() > 0 ) ) {
-			$result = $this->parser->recursiveTagParseFully( $this->mIntro ) . $result;
+			if ( $outputmode == SMW_OUTPUT_HTML && $this->isHTML ) {
+				$result = Message::get( [ 'smw-parse', $this->mIntro ], Message::PARSE ) . $result;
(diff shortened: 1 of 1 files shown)


YvarRavy:
could you check this out? @gesinn-it-ilm

gesinn-it-ilm:
@JeroenDeDauw could you please review this PR? Thanks.

JeroenDeDauw:
Costs too many brain cycles. If another dev vouches for this code I am happy to merge

thomas-topway-it:
I will check it soon on my side, I was also planning to do a few updates to carousel format

thomas-topway-it:
I'm sorry for the delay. As pointed out [here](https://github.com/SemanticMediaWiki/SemanticResultFormats/issues/853) the error shows when Datatables is called from the **Special:Ask** page and `$this->mOptions` are not set in the MW `Parser` (when `$this->parser->recursiveTagParseFully` is called from the function `handleNonFileResult`) 

Therefore

`$isValid = $this->parser->getOptions() && $this->parser->getTitle();`

as in `SMW\Parser\RecursiveTextProcessor` could be checked before using `$this->parser->recursiveTagParseFully` in order to set Parser's `setOptions` explicitly, or just to use the original solution (from `SMW\Query\ResultPrinters\ResultPrinter`) pointed out by you.

(I will merge it, then I will add a comment in the code with instructions to fix the original Datatables solution if needed)

thomas-topway-it (approved):
assessment & tested

PR #861, opened 2024-11-12
handle SPARQLStore backends

simontaurus:
fixes #827 
See also https://github.com/SemanticMediaWiki/SemanticMediaWiki/issues/5748

With the next SMW release including https://github.com/SemanticMediaWiki/SemanticMediaWiki/commit/8bfd7a96a40d981bfd1eb0e2917d76f1543a0990 the Closure can be removed

PR #865, opened 2024-11-23
Use orthogonal data for DataTables

### LINKED ISSUE #831: [datatables] Datatables should sort Quantity data numerically
### Setup

- MW version: 1.39.7
- DB (MySQL etc.): 10.11.7-MariaDB-1:10.11.7+maria~ubu2204
- PHP version: 8.1.28 (fpm-fcgi)
- SMW version: 4.1.3
- SRF version: 4.2.1

### Issue

`Quantity` type properties are sorted alphabetically instead of numerically in datatable, which is unexpected by the users.

**Steps to reproduce**
1. Set up an inline ask query on a page
2. Set format set as 'datatables'
3. Add any properties with the `Quantity` type
4. Save the page
5. Click on the property in the table header to sort the column

**Expected result:**
The datatable should be sorted by that property numerically.

**Observed result:** 
The datatable is sorted by that property alphabetically. 

Example page with the issue: https://starcitizen.tools/Gun
The `Damage`, `Range`, and `Muzzle velocity` properties have the `Quantity` type.

alistair3149:
This will make DataTables easier to expand and maintain. Now we can assign different attributes to the data, and this will resolve any sorting and filtering issues that we had.

### Added
Instead of assigning HTML directly as data, we can assign an object with the following keys:
- `display` - The display value of the data (same as what we had)
- `filter` - Value used for filtering (e.g. SearchBuilder)
- `sort` - Value used for sorting

### Removed
- `any-number` - Not needed anymore because sort key are always sortable
- `datatables-column.type` - Same reason as above

https://datatables.net/manual/data/orthogonal-data

Fix #831


JeroenDeDauw:
Looks good. @thomas-topway-it can you also check this?

malberts:
The code change looks fine, although I didn't do any manual tests.

Are there any possible backward compatibility issues here? Since this changes what the PHP outputs, I imagine pages with outdated caches will serve the old HTML with the new Javascript and then break. (Again, not manually confirmed, just checking.)

alistair3149:
> The code change looks fine, although I didn't do any manual tests.
> 
> Are there any possible backward compatibility issues here? Since this changes what the PHP outputs, I imagine pages with outdated caches will serve the old HTML with the new Javascript and then break. (Again, not manually confirmed, just checking.)

It would not be backward compatible and require a parser purge on the affected pages.

thomas-topway-it:
for some reason I've missed this pull request until today. I've still to test it by the way it is great that you are moving to use orthogonal data (which as far as I understand are suitable to display structured data, like sub-objects in the context of SMW).

Regarding the sorting enhancement it seems an optimal solution, I will test with one of our demo (with the sorting relying on any-number to double check how it works)

@malberts you are correct, the old html may break the new javascript, however I don't see this as a "backward compatibility issue", just the requirement to update the pages

JeroenDeDauw:
@thomas-topway-it so you will test this soonish?

thomas-topway-it:
@JeroenDeDauw I cannot access the demo server currently, however it looks a cutting-edge solution, so if there is any issue I or @alistair3149 can fix it afterwards

thomas-topway-it:
this can also be merged https://github.com/SemanticMediaWiki/SemanticResultFormats/compare/master...OpenSemanticLab:SemanticResultFormats:dev?expand=1

but I guess it should be rebased @simontaurus

simontaurus:
> Is it the same as this PR? https://github.com/SemanticMediaWiki/SemanticResultFormats/pull/861

https://github.com/SemanticMediaWiki/SemanticResultFormats/compare/master...OpenSemanticLab:SemanticResultFormats:dev?expand=1
contains both
https://github.com/Knowledge-Wiki/SemanticResultFormats/tree/datatables-fix-template-parameter-ajax
and
https://github.com/simontaurus/SemanticResultFormats/tree/patch-1

alistair3149:
@thomas-topway-it would you be able to review the PR?

thomas-topway-it:
@alistair3149 I'm testing, a few minutes ... I have seen the new theme, is really beautiful but shouldn't be enabled via a parameter ?

thomas-topway-it:
I have tested and think I understood how it works. Could we add 

```
		if ( !count( $dataValues ) ) {
			return [
				'display' => $html,
				'filter' => '',
				'sort' => ''
			];
		}
```

or something similar before 

`		$sortKey = $dataValues[0]->getDataItem()->getSortKey();`

since `$dataValues` could be empty (it triggered an error on my test article)

Also, the idea is really clever, and of course the best solution so I totally agree to commit it, however how pseudo numeric titles are handled, like :

-- 1_Some title
-- 2_Some title
etc. ?

I think there was something similar in one of our demos ...

alistair3149:
> @alistair3149 I'm testing, a few minutes ... I have seen the new theme, is really beautiful but shouldn't be enabled via a parameter ?

We can either add a MW config key to disable that globally or/and add a datatable parameter to disable that . We did some modifications on both the styles and script to build the UI so we might have to refactor the changes together into the same place.

alistair3149:
> since $dataValues could be empty (it triggered an error on my test article)

It should be handled now. Would you be able to test it again?

> however how pseudo numeric titles are handled, like :
-- 1_Some title
-- 2_Some title
etc. ?
I think there was something similar in one of our demos ...

I'm not quite grasping the concept. Would you mind elaborating further?

thomas-topway-it (approved):
I have tested, thanks a lot

PR #864, opened 2024-11-21
Bump DataTables from 1.13.2 to 2.1.8

alistair3149:
The existing features seem to work fine. It is more maintainable and easier to develop using the v2 version.

PR #867, opened 2024-11-27
Use layout instead of dom for datatables

alistair3149:
Follow up to #864

### DIFF
--- a/formats/datatables/DataTables.php
+++ b/formats/datatables/DataTables.php
@@ -241,12 +241,6 @@ class DataTables extends ResultPrinter {
 			'default' => '',
 		];
 
-		$params['datatables-dom'] = [
-			'type' => 'string',
-			'message' => 'srf-paramdesc-datatables-library-option',
-			'default' => 'lfrtip',
-		];
-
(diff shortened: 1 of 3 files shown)


thomas-topway-it:
@alistair3149 it seems really fine! Just a note: will `top9End` be placed on the right if `top9Start` is missing ?

alistair3149:
@thomas-topway-it yes it will place on the end even if start is missing.
I put it there for now because I am not sure how it is used normally.

On a side note, do you have more example pages on how DataTables are being used currently in SMW?  I am planning to style the DataTables closer to the [Table component in Codex](https://doc.wikimedia.org/codex/main/components/demos/table.html), and it would be helpful to be able to take different layouts into consideration.

thomas-topway-it:
I see Codex has improved a lot lately ...
regarding examples I have only this
https://www.semantic-mediawiki.org/wiki/Help:Datatables_format#Demos
this
https://starcitizen.tools/Gun
and this
https://wikincat.org/wiki/Form:Circulation

jaideraf:
So, this is ready to be merged, right?

PR #868, opened 2024-11-27
Support showing all rows as a length option in DataTables

alistair3149:
By adding `-1` to `datatables-lengthMenu`, all rows can be shown.

image

### DIFF
--- a/formats/datatables/resources/ext.srf.formats.datatables.js
+++ b/formats/datatables/resources/ext.srf.formats.datatables.js
@@ -558,6 +558,14 @@
 				});
 			}
 
+			// Replace -1 in lengthMenu with 'all' label
+			var showAll = options.lengthMenu.indexOf( -1 );
+			if ( showAll !== -1 ) {
+				var labels = options.lengthMenu.slice();
+				labels[showAll] = mw.msg( 'allmessages-filter-all' ); // stealing MW core messages :D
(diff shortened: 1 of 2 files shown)


formats/datatables/resources/ext.srf.formats.datatables.js
```diff
@@ -558,6 +558,14 @@
 				});
 			}
 
+			// Replace -1 in lengthMenu with 'all' label
+			var showAll = options.lengthMenu.indexOf( -1 );
+			if ( showAll !== -1 ) {
+				var labels = options.lengthMenu.slice();
```
JeroenDeDauw, line 564: Why the `slice()`?

alistair3149: Need to clone the array so that the original `options.lengthMenu` wouldn't be altered

JeroenDeDauw: > so that the original options.lengthMenu wouldn't be altered.

Why?

alistair3149: On Line 566, we set two arrays to the `lengthMenu`. The first array is for the numerical values (`-1` represents `all`, which is the original values), the second array is for the label (`all` represents `-1`).

Issue #614, opened 2020-08-27
[Incoming] unnecessary div returned preventing further reuse of results


IRA1777:
### Setup
- SRF version: 3.1.0

### Issue
Incoming format returns unnecessary div that prevent further data processing.
getFormatOutput function add it in the "Beautify class selector" section.

**Steps to reproduce**
Put a query result of format "incoming" in a variable.
Try to manipulate the variable: it never has the expected behaviour -> returned value contains html elements.

PR #870, opened 2024-12-07
Fix for #614

D-Groenewegen:
Fix for #614. When no template is used, removes the html (a div element with the 'srf-incoming' class) from the result so that it allows for further reuse.

### DIFF
--- a/formats/incoming/SRF_Incoming.php
+++ b/formats/incoming/SRF_Incoming.php
@@ -141,19 +141,24 @@ class SRFIncoming extends ResultPrinter {
 			$result = implode( $this->params['sep'], array_keys( $data ) );
 		}
 
-		// Beautify class selector
-		$class = $this->params['template'] ? 'srf-incoming ' . str_replace(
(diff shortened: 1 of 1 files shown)


gesinn-it-gea:
Thanks for your PR! At the moment, the changed file is not covered by any tests: https://app.codecov.io/gh/SemanticMediaWiki/SemanticResultFormats/blob/master/formats%2Fincoming%2FSRF_Incoming.php. Do you think you can add some testing?

D-Groenewegen:
Apologies for the delay - I was hit with the man flu shortly afterwards. I can't currently find the time to add a test, but when I do, I can give it a try. Any guide we should be following?

gesinn-it-gea:
@D-Groenewegen ...take a look at the existing tests, maybe start with the JSONScript tests. There are also JSONScript tests in SMW itself. See also: https://github.com/SemanticMediaWiki/SemanticMediaWiki/blob/master/tests/phpunit/Integration/JSONScript/README.md

PR #871, opened 2024-12-08
Implement basic Codex styles to DataTables

alistair3149:
- Import Codex tokens like in https://github.com/SemanticMediaWiki/SemanticMediaWiki/pull/5786
- Add basic Codex styles to DataTables

https://github.com/user-attachments/assets/23247eb1-3a17-49bb-ad88-32515bbd5164

I will make a follow-up patch to clean up the icons


formats/datatables/resources/ext.srf.formats.datatables.js
```diff
@@ -550,11 +560,19 @@
...
+			var lengthMenuLabels = options.lengthMenu.slice();
 			if ( showAll !== -1 ) {
-				var labels = options.lengthMenu.slice();
-				labels[showAll] = mw.msg( 'allmessages-filter-all' ); // stealing MW core messages :D
-				options.lengthMenu = [ options.lengthMenu, labels ];
+				lengthMenuLabels[showAll] = mw.msg( 'srf-ui-datatables-label-rows-all' );
+			}
+			// Format value into readable label
+			for (var i = 0; i < lengthMenuLabels.length; i++) {
+				console.log( typeof lengthMenuLabels[i] );
```
malberts, line 569: This debugging can be removed.

JeroenDeDauw:
@thomas-topway-it could you have a look at this?

formats/datatables/resources/ext.srf.formats.datatables.js
```diff
@@ -550,11 +560,18 @@
...
 			var showAll = options.lengthMenu.indexOf( -1 );
+			var lengthMenuLabels = options.lengthMenu.slice();
 			if ( showAll !== -1 ) {
-				var labels = options.lengthMenu.slice();
-				labels[showAll] = mw.msg( 'allmessages-filter-all' ); // stealing MW core messages :D
-				options.lengthMenu = [ options.lengthMenu, labels ];
+				lengthMenuLabels[showAll] = mw.msg( 'srf-ui-datatables-label-rows-all' );
+			}
+			// Format value into readable label
+			for (var i = 0; i < lengthMenuLabels.length; i++) {
```
JeroenDeDauw, line 568: There is some inconsistent code style spacing in the new code

alistair3149: It was my bad, I was used to the spacing scheme in other MW projects.
Would it be helpful to set up eslint and stylelint for the SRF down the line? There would probably be a lot of violations but it would be great to enforce some consistency in code styling.

JeroenDeDauw: Oh, the code here does not use the normal MW style? I would just use the standard MW style. The code here comes from all kinds of random contributors, so not surprising it deviates at places.

Yes, having some CS rules could help... though I wonder how much they would get in the way here. Having the option to merge something slightly inconsistent is nice, because repeatedly harassing new contributors over spaces isn't likely to endear them to the project. So _shrug_

alistair3149: Almost all the JS that I worked on within SRF isn't using standard code styles. It might be safe to assume that majority of the code do not conform to the styles based on the age of the code themselves.

I am a bit torn about this because if we are to enable CS rules, the CI will light up like a Christmas tree... I don't have the bandwidth to do it at the moment but perhaps some day in the future.

PR #875, opened 2024-12-17
Fix CI

paladox:
Since [0] It runs composer test and composer phpunit meaning right now it is running phpunit twice.

We cleanup composer a bit.

Also use the anonymous git clone url and not the ssh one for the submodule.

[0] https://github.com/gesinn-it-pub/docker-compose-ci/commit/08783a070080fa0bb519508dd7b1c81fd9c78f88


composer.json
```diff
@@ -83,8 +83,7 @@
 	},
 	"scripts":{
 		"test": [
-			"@analyze",
-			"@phpunit"
```
JeroenDeDauw, line 87: This also changes the behavior for people that run `composer test` locally. I suspect the issue you are fixing comes from the GH Actions config

paladox: Yup. Unless https://github.com/gesinn-it-pub/docker-compose-ci/commit/08783a070080fa0bb519508dd7b1c81fd9c78f88 gets reverted.

paladox: The readme specifies > This extension provides unit and integration tests and is usually run by a [continues integration platform](https://docs.github.com/en/actions) but can also be executed locally using the shortcut command composer phpunit from the extension base directory.

Anyways. This unblocks the CI, so either this should be merged or https://github.com/gesinn-it-pub/docker-compose-ci/commit/08783a070080fa0bb519508dd7b1c81fd9c78f88 reverted.

gesinn-it-gea: I think we should revert https://github.com/gesinn-it-pub/docker-compose-ci/commit/08783a070080fa0bb519508dd7b1c81fd9c78f88

gesinn-it-gea: This is mainly because many other extensions are using docker-compose-ci. Forcing them to change their compose setup is not ideal.

PR #879, opened 2024-12-18
UI improvements to dataTables

alistair3149:
- Make searchBuilder and searchPanes title more consistent
- Redesign search pane layout
- Change pagination buttons into icon buttons
- Use Codex icons when applicable
- Update button styles to match with current Codex standards
- Add i18n for searchBuilder and searchPanes title
- Update token.less to Codex Design Tokens v1.18.0 and polyfill new LESS variables

https://github.com/user-attachments/assets/73e6b581-9c83-4136-8fe0-5a4a2616a711


resources/tokens.less
```diff
@@ -333,6 +342,29 @@
...
 @spacing-typeahead-search-focus-addition: calc( ( @spacing-start-typeahead-search-figure + @min-size-search-figure ) - ( @min-size-icon-medium + @spacing-50 ) ); // The amount the width of the input increases when it is focused to allow for the extra spacing around the search figures. The caret position should remain static. This calculates the padding-left of the input when expanded minus the padding-left of the input when not expanded. (Note that both padding values actually include `@spacing-50` as well, but it was left out of the calculation for simplicity's sake.)
 
+/**
+ * Codex polyfill for MW 1.43 / Codex Design Tokens v1.14.0
+ * This section is used to polyfill Codex tokens introduced since MW 1.43,
+ * as dark mode is introduced with Vector-2022 in MW 1.43.
+ *
+ * TODO: Figure out a way to polyfill tokens for dark mode
+ */
+/** @since v1.16.0 */
```
JeroenDeDauw, line 352: Does this have to be in the same file? I'm worried that having this section at the bottom is easy to overlook when updating the preceding copied codex token code.

JeroenDeDauw: Then again, maybe we never need to update it? Can we drop these copies once we require MW 43 or later?

alistair3149: > Does this have to be in the same file? I'm worried that having this section at the bottom is easy to overlook when updating the preceding copied codex token code.

It can be in a separate LESS file and imported through `token.less`. Will reorganize it in the same PR.

> Then again, maybe we never need to update it? Can we drop these copies once we require MW 43 or later?

I would suggest keeping a copy so that we can keep compatibility across MW versions. Design systems like Codex are often an evolving standard (and it did move much quicker than OOUI) and each MW version bundles with a different version of Codex. There are many fixes and improvements that are useful in the newer version of Codex, and having the newer version of Codex means that we can decouple UI development from the oldest Codex/MW version that we have to support.

alistair3149: > It can be in a separate LESS file and imported through token.less. Will reorganize it in the same PR.

Done. I have separated the tokens into three files.
- `tokens.less` serves as the entry point that is used to be imported by other LESS files if needed, or any other SRF tokens in the future.
- `tokens-codex.less` is a copy of the current Codex Design Tokens
- `tokens-codex-polyfill.less` contains the polyfill needed for Codex since MW 1.43.

PR #887, opened 2024-12-19
DO NOT MERGE

paladox:


paladox:
It says it's being killed and I don't know why, I'm going to have a guess that it's memory and OOM'ing?

PR #892, opened 2024-12-27
Add tests for MW 1.41

paladox:


paladox:
Seems this was based on an old version, done https://github.com/SemanticMediaWiki/SemanticResultFormats/pull/894 instead. But same problem. It's getting killed.

PR #896, opened 2024-12-29
Add CI for MW 1.42/43

paladox:


paladox:
Any ideas on how to fix the error @gesinn-it-gea / @gesinn-it-ilm ?

paladox:
I've added some skips to the test for now, but they'll need to be fixed in the future. (I just don't know how to fix them or why they are starting to fail). It could mean either the fix is needed in the test or that the functionality is actually broken in MW 1.42, I just don't know.

PR #900, opened 2025-01-08
Require MW 1.39+ and at least SMW 4.2+

paladox:


JeroenDeDauw:
Is this necessary? A bunch of improvements and fixes have been made, so it'd be nice to have another release for those on MW 35 with these

alistair3149:
We actually have to use SMW 5.0 since ef1c22f9fff96ab40ee7f1cd16ae4f2424bef5c0 because of the ResourceLoader module changes. 

We can add a version check for SMW for that change specifically to keep backward compatibility.

Other than that the DataTables changes might work in 1.35 since we backported all the Codex design tokens, but it would have to be tested

JeroenDeDauw:
If we have to jump through these extra hoops, then that is a good reason to raise the minimum requirements.

PR #903, opened 2025-01-29
Test for #614

D-Groenewegen:
Adds a JSONScript test file for the incoming format, focusing for now on #614.


D-Groenewegen:
Interestingly, `({{#ask: [[Example/SRF0614/D.1]] |mainlabel=- |link=none |format=incoming |limit=999 |sep=; }})`, where `Example/SRF0614/D.1` receives incoming properties from two different pages, produces an empty an empty string.

```
==== (actual) ====
<div class="mw-content-ltr mw-parser-output" lang="en" dir="ltr"><p>()
</p></div>
==== (StringContains expected) ====
[ (Is related to;Not to be confused with) ]

Failed asserting that 0 matches expected 1.

/var/www/html/extensions/SemanticMediaWiki/tests/phpunit/Utils/Validators/StringValidator.php:76
/var/www/html/extensions/SemanticMediaWiki/tests/phpunit/Utils/Validators/StringValidator.php:31
/var/www/html/extensions/SemanticMediaWiki/tests/phpunit/Utils/JSONScript/QueryTestCaseProcessor.php:241
/var/www/html/extensions/SemanticMediaWiki/tests/phpunit/JSONScriptServicesTestCaseRunner.php:503
/var/www/html/extensions/SemanticMediaWiki/tests/phpunit/JSONScriptServicesTestCaseRunner.php:442
/var/www/html/extensions/SemanticMediaWiki/tests/phpunit/JSONScriptServicesTestCaseRunner.php:144
/var/www/html/extensions/SemanticMediaWiki/tests/phpunit/JSONScriptTestCaseRunner.php:196
```

A previous version, without `mainlabel=-` and `link=none`, returns the page. Something tells me the `incoming` format is not enabled on the test wiki. Will now set `srfgFormats` explicitly.

D-Groenewegen:
I give up. @gesinn-it-gea, I was hoping you'd be able to tell what's wrong with the integration test. It looks like either the incoming format refuses to be enabled, in spite of the explicit settings, or there may be other reasons why the query is simply not able to pick up incoming semantic data.

[Edit - 9 July] Looks like the solution will be #1117, which needs to be merged first.

PR #904, opened 2025-02-02
Datatables redesign client side cache mechanism

thomas-topway-it:
* removes parameter `defer-each` (set default limit to 100 and retrieve all rows until the limit both on page load and via Ajax)
* redesign client-side cache mechanism to correctly handle non-subsequent pages and to take advantage from the retrieved rows against table page length

(test further before merging)


alistair3149:
Will look into it after #908

alistair3149 (approved):
Tested working locally

thomas-topway-it:
@alistair3149 thanks a lot for testing, this was an important update !

PR #908, opened 2025-02-26
Add missing use statement and deprecate old methods

alistair3149:


paladox:
Needs either manually fixing or to run composer phpcs-fix

PR #928, opened 2025-03-28
Prepare 5.0.0 release notes

JeroenDeDauw:


JeroenDeDauw:
@alistair3149 could you replace the links under the DataTables point with useful descriptions? Does not need to stay a list with the current number of elements. I imagine some PRs do more than one noteworthy thing. (I'm asking this as SRF maintainer.)

PR #945, opened 2025-06-16
Fix template expansion to allow spaces in template names

HamishSlater:
Strict whitespace handling was recently introduced with #931

Trim to allow e.g. `|+template = mytemplate`

Alternatively, we could embrace the strictness, however docs should be updated:

https://www.semantic-mediawiki.org/wiki/Help:Datatables_format#Format_specific


HamishSlater:
#944 945

Issue #910, opened 2025-02-26
DataTables: Table name component contains unexpected quote or dot character with SearchPane


alistair3149:
Encountered the following error when attempting to enable search pane for some of the datatables.
Live demo: https://starcitizen.tools/List_of_pledge_vehicles

---

**Code:** `internal_api_error_DBLanguageError`

**Information:** `[4179acf03d61be85a0234530] Exception caught: Table name component contains unexpected quote or dot character`

**Error Class:** `Wikimedia\Rdbms\DBLanguageError`

**Trace:**
```
Wikimedia\Rdbms\DBLanguageError at /var/www/mediawiki/includes/libs/rdbms/platform/SQLPlatform.php(1154)
from /var/www/mediawiki/includes/libs/rdbms/platform/SQLPlatform.php(1154)

#0 /var/www/mediawiki/includes/libs/rdbms/platform/SQLPlatform.php(1094): Wikimedia\Rdbms\Platform\SQLPlatform->extractTableNameComponents(string)
#1 /var/www/mediawiki/includes/libs/rdbms/platform/SQLPlatform.php(1060): Wikimedia\Rdbms\Platform\SQLPlatform->qualifiedTableComponents(string)
#2 /var/www/mediawiki/includes/libs/rdbms/platform/SQLPlatform.php(1023): Wikimedia\Rdbms\Platform\SQLPlatform->tableName(string)
#3 /var/www/mediawiki/includes/libs/rdbms/platform/SQLPlatform.php(924): Wikimedia\Rdbms\Platform\SQLPlatform->tableNameWithAlias(string, string)
#4 /var/www/mediawiki/includes/libs/rdbms/platform/SQLPlatform.php(711): Wikimedia\Rdbms\Platform\SQLPlatform->tableNamesWithIndexClauseOrJOIN(array, array, array, array)
#5 /var/www/mediawiki/includes/libs/rdbms/database/Database.php(3305): Wikimedia\Rdbms\Platform\SQLPlatform->selectSQLText(array, string, string, string, array, array)
#6 /var/www/mediawiki/includes/libs/rdbms/database/DatabaseMySQL.php(644): Wikimedia\Rdbms\Database->selectSQLText(string, string, string, string, array, array)
#7 /var/www/mediawiki/includes/libs/rdbms/database/Database.php(1337): Wikimedia\Rdbms\DatabaseMySQL->selectSQLText(string, string, string, string, array, array)
[...]

PR #942, opened 2025-05-30
Fix SearchPanes

HamishSlater:
Fixes #910

PR #961, opened 2025-08-23
use structured queries

thomas-topway-it:
-- considers this https://github.com/SemanticMediaWiki/SemanticResultFormats/pull/942

-- fixes https://github.com/SemanticMediaWiki/SemanticResultFormats/issues/910

-- uses structured queries instead of raw joins in the from clause

### DIFF
--- a/formats/datatables/SearchPanes.php
+++ b/formats/datatables/SearchPanes.php
@@ -70,6 +70,92 @@ class SearchPanes {
 		return $this->searchPanesLog;
 	}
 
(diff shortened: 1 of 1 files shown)


alistair3149:
Merging it to a separate branch for testing

thomas-topway-it:
@alistair3149 done -- I wanted to test also here https://fina.knowledge.wiki/Test_Datatables but there is some other issue.
After testing, I have a better idea how to use the QuerySegmentListProcessor, but the functionality will be the same

Issue #962, opened 2025-09-02
[Filtered] Fix call to undefined method WikiPage::factory


JeroenDeDauw:
Seen on MediaWiki 1.43

> Call to undefined method WikiPage::factory()

Source: https://github.com/SemanticMediaWiki/SemanticResultFormats/blob/ca6b8588896ae04751e8741ec14881844ce55b38/formats/filtered/src/View/MapView.php#L308

PR #963, opened 2025-09-02
Stop using removed WikiPage::factory

JeroenDeDauw:
For 
* https://github.com/SemanticMediaWiki/SemanticResultFormats/issues/962

Untested


malberts:
Tested that the page no longer throws an exception.

PR #968, opened 2025-09-24
Use structured queries for datatables

alistair3149:
Created from #961

> -- considers this https://github.com/SemanticMediaWiki/SemanticResultFormats/pull/942
-- fixes https://github.com/SemanticMediaWiki/SemanticResultFormats/issues/910
-- uses structured queries instead of raw joins in the from clause


thomas-topway-it:
@alistair3149  @simontaurus have you been able to test this ?

PR #975, opened 2025-10-10
Fix tableName() usage in SearchPanes

huaj1ng:
* Solves https://github.com/SemanticMediaWiki/SemanticResultFormats/issues/910 with reference to https://github.com/SemanticMediaWiki/SemanticResultFormats/pull/942 
* In select queries, we use $qobj->fromTables() and $qobj->joinConditions added in https://github.com/SemanticMediaWiki/SemanticMediaWiki/pull/5715


thomas-topway-it:
@huaj1ng thanks this is the proper solution after this change https://github.com/SemanticMediaWiki/SemanticMediaWiki/pull/5715 

could you please rebase over this https://github.com/SemanticMediaWiki/SemanticResultFormats/pull/986  adding a condition to use your patch with the proper SMW version ? then I will try to simplify the cloned `QuerySegmentListProcessor.php` class for backwards compatibility

thomas-topway-it (changes requested):
see comment

huaj1ng:
Please excuse the delayed reply. After checking SemanticMediaWiki/SemanticResultFormats#986 carefully, I noticed that it already interprets all key ideas of this PR in a neater way, which looks cleaner than the style of SemanticMediaWiki/SemanticMediaWiki#5715, the one I was trying to mimic.

From what I can see, the purpose of this PR is then drained - what I can improve is now merely a few null checks - it is up to your decision to use them or not.

I am not sure about "adding a condition to use your patch with the proper SMW version": as for the `tableName()` usage we are only using a similar way of constructing queries like SemanticMediaWiki here, but not relying on certain SMW version - please share your insight if my understanding is incomplete.

thomas-topway-it (approved):

thomas-topway-it:
@huaj1ng ok, thanks a lot. I mean that the `SRF\DataTables\QuerySegmentListProcessor` is not needed after this change https://github.com/SemanticMediaWiki/SemanticMediaWiki/pull/5715  so the class  `SRF\DataTables\QuerySegmentListProcessor`  could be used conditionally depending the SMW version. 
It's not urgent as long as the original `QuerySegmentListProcessor` does not change further.

PR #989, opened 2026-01-28
Fixed issue where using DV Time there would be a fatal error in an ask query

### LINKED ISSUE #988: Error `Call to undefined method SMWTimeValue::getDisplayTitle()` when using `Modification date::+` and result format graph
When using a Semantic MediaWiki ASK query with a result format set to `graph`, I encounter the following fatal error:
```Error: Call to undefined method SMWTimeValue::getDisplayTitle()
Backtrace:
from /path/to/SemanticResultFormats/src/Graph/GraphPrinter.php(220)
#0 /path/to/SemanticResultFormats/src/Graph/GraphPrinter.php(155): SRF\Graph\GraphPrinter->processResultRow(array)```

YvarRavy:
Fixes https://github.com/SemanticMediaWiki/SemanticResultFormats/issues/988

### DIFF
--- a/src/Graph/GraphPrinter.php
+++ b/src/Graph/GraphPrinter.php
@@ -214,10 +214,10 @@ class GraphPrinter extends ResultPrinter {
 
 			while ( ( $object = $result_array->getNextDataValue() ) !== false ) {
 				$hasProperty = $object->getProperty();
-				if ( $object instanceof \SMW\DataValues\StringValue ) {
-					$objectText = $object->getShortWikiText();
-				} else {
+				if ( method_exists( $object, 'getDisplayTitle' ) ) {
 					$objectText = $object->getDisplayTitle();
+				} else {
+					$objectText = $object->getShortWikiText();
 				}
 
 				$includeAsEdge = !$showGraphFields || $isPageType || $request->isMode( PrintRequest::PRINT_CHAIN );


paladox:
Merge conflicts need resolving.

gesinn-it-gea:
Thanks for the fix, @YvarRavy! Unfortunately this PR has become outdated — around two weeks after it was opened, the same code path in `GraphPrinter::processResultRow()` was independently reworked by commits [e53a40d3](https://github.com/SemanticMediaWiki/SemanticResultFormats/commit/e53a40d374f394e52379c4d7cce2617b8262cbb0) and [be6ad81f](https://github.com/SemanticMediaWiki/SemanticResultFormats/commit/be6ad81fa445d5977d5649b64914a98b68e01a22), which replaced the `StringValue`/`getDisplayTitle()` branching with an `$isPageType` gate and a `getWikiValue()` fallback — a more robust fix, since it doesn't rely on `method_exists()` matching by coincidence for non-page types. GitHub now reports this PR as conflicting with `master`.

The underlying bug is fixed on `master` and covered by a regression test (`GraphPrinterTest::testProcessResultRowUsesWikiValueForNonPageType`) for exactly the `SMWTimeValue`/`Modification date::+` scenario described in #988. Closing in favor of that fix — see #988 for details.

PR #996, opened 2026-02-27
ksort() causes invalid SQL JOIN order in SearchPanes with 5+ OR values

YvarRavy:
this fixes #995


thomas-topway-it (approved):
I am sorry for the delay, thanks a lot for the explanation here https://github.com/SemanticMediaWiki/SemanticResultFormats/issues/995

PR #998, opened 2026-04-08
Add explicit return types and Linker import in TreeResultPrinter for SMW compatibility

squirrel-work:
This updates TreeResultPrinter to match typed method signatures now present in SMW parent classes. Closes #6581

Changes:
- add string return type to getName()
- add array return type to getParamDefinitions()
- add void return type to postProcessParameters()
- add ?Linker return type to getLinker()
- add use MediaWiki\Linker\Linker

Why:
Without these changes, PHP fatal errors occur because TreeResultPrinter overrides no longer match the corresponding SMW parent method signatures.

Scope:
This is intended as a compatibility fix only. No tree-format logic was changed.

### DIFF
--- a/formats/tree/TreeResultPrinter.php
+++ b/formats/tree/TreeResultPrinter.php
@@ -10,6 +10,7 @@ namespace SRF\Formats\Tree;
 
 use Exception;
 use Html;
+use MediaWiki\Linker\Linker;
 use SMW\DIProperty;
 use SMW\DIWikiPage;
 use SMW\Query\QueryResult;
@@ -37,7 +38,7 @@ class TreeResultPrinter extends ListResultPrinter {
 	 * (non-PHPdoc)
 	 * @see \SMW\Query\ResultPrinters\ResultPrinter::getName()
 	 */
-	public function getName() {
+	public function getName(): string {
 		// Give grep a chance to find the usages:
 		// srf-printername-tree, srf-printername-ultree, srf-printername-oltree
 		return \Message::newFromKey( 'srf-printername-' . $this->mFormat )->text();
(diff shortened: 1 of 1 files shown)


jaideraf:
CI:
```
There was 1 error:

1) SRF\Tests\Integration\JSONScript\JsonTestCaseScriptRunnerTest::testCaseFile 
with data set "tree-07.json" ('/var/www/html/extensions/Sema...7.json')
TypeError: SRF\Formats\Tree\TreeResultPrinter::getLinker(): 
Return value must be of type ?MediaWiki\Linker\Linker, Linker returned
```

formats/tree/TreeResultPrinter.php
```diff
@@ -258,7 +259,7 @@ protected function getHashOfNodes() {
 	 *
 	 * @return \Linker
 	 */
-	public function getLinker( $firstcol = false ) {
+	public function getLinker( $firstcol = false ): ?Linker {
```
JeroenDeDauw, line 262: You can remove the `@return` doc as it no longer adds value and also says something different (not-nullable)

JeroenDeDauw:
> Return value must be of type ?MediaWiki\Linker\Linker, Linker returned

The test might be the issue. Mayve its creating a mock for `\Linker`, which might no longer exist, not even as alias. I expect a test to need updating.

alistair3149:
> > Return value must be of type ?MediaWiki\Linker\Linker, Linker returned
> 
> The test might be the issue. Mayve its creating a mock for `\Linker`, which might no longer exist, not even as alias. I expect a test to need updating.

It's also caused by the MW upstream namespace class change. We can do alias but it'll be messy, or we can drop 1.39 support to sync with SMW.

JeroenDeDauw:
Can totally drop support for MW < 1.43

paladox:
Done in https://github.com/SemanticMediaWiki/SemanticResultFormats/pull/1001 instead

paladox:
https://github.com/SemanticMediaWiki/SemanticResultFormats/pull/1001 has been merged

PR #1002, opened 2026-04-11
Bump version to 6.0.0

paladox:
With the droppage of MW 1.39 and the requirement of at least MW 1.43+ now. Let's bump the version to 6.0.0.

### DIFF
--- a/extension.json
+++ b/extension.json
@@ -1,6 +1,6 @@
 {
     "name": "SemanticResultFormats",
-	"version": "5.2.0",
+	"version": "6.0.0",
 	"author": [
 		"James Hong Kong",
 		"Stephan Gambke",


kghbln:
@paladox @paladox Are we ready to make the release and tag 6.0.0? I could do this.

paladox:
> @paladox @paladox Are we ready to make the release and tag 6.0.0? I could do this.

Yes I think so.

thomas-topway-it:
there are also these
-- https://github.com/SemanticMediaWiki/SemanticResultFormats/pull/999
-- https://github.com/SemanticMediaWiki/SemanticResultFormats/pull/1000

the first is not urgent but the second restores this https://github.com/SemanticMediaWiki/SemanticResultFormats/pull/976  which otherwise causes issues with filtered's map

thomas-topway-it:
btw `ext.srf.filtered.leaflet.js`, simply won't be loaded, there are no changes in this, so  possibly it's not urgent either

paladox:
https://github.com/SemanticMediaWiki/SemanticResultFormats/pull/999 should be rebased so we can see whether it passes under MW 1.43-45

PR #1037, opened 2026-05-09
update datatables - Listwidget prefix parameter

thomas-topway-it:
- fixes ListWidget prefix
- refactor datatables getCellContent
-  updates Datatables (use MW connection since https://github.com/SemanticMediaWiki/SemanticMediaWiki/commit/cb5061809fd3937ab2f1fb823e6ea981b9ca1380)


krabina:
Do not integrate this yes. I get
`TypeError: Cannot assign SMWQuery to property SRF\DataTables::$query of type ?SMW\Query\Query`
when running refreshData.php

alistair3149:
> Do not integrate this yes. I get `TypeError: Cannot assign SMWQuery to property SRF\DataTables::$query of type ?SMW\Query\Query` when running refreshData.php

That only happens when SMW 7 is not used, because of the namespace class changes. It doesn't play well with strict PHP native types.

Perhaps it is more manageable to only target one SMW major version and bump SRF to 7.0.0-alpha like [SemanticScribunto](https://github.com/SemanticMediaWiki/SemanticScribunto/blob/7acfaa1e10a5836af1762aa8ef06658b9ef0c292/extension.json#L3).

thomas-topway-it:
this PR is actually coupled with https://github.com/SemanticMediaWiki/SemanticMediaWiki/pull/6307 so I will try to finalize this after the latter is merged

PR #1047, opened 2026-06-10
Fix result format name message keys for statistics and carousel

### LINKED ISSUE #1043: Some names of result formats are not properly showing in Special:Ask
* SMW: 7.0.0-alpha (18e19b7)
* SRF: 6.0.0 (54c78a6)
* MW: 1.43.8
* PHP: 8.1.32 

Some names of result formats are not properly showing in `Special:Ask`. See:

https://github.com/user-attachments/assets/ba4997a5-578f-40b1-9226-fb349ee1b5a6

kghbln:
The statistical math formats (variance, mode, range, the quartile family, ...) were added to the i18n files with hyphenated message keys, but SRF_Math.php builds the name key with underscores (srf_printername_<format>), so Special:Ask showed the raw keys. The carousel format has the opposite mismatch: Carousel.php requests the hyphenated key while the i18n files carried the underscored one.

Rename the affected keys in en.json and qqq.json to match what the code requests, and give the statistical formats proper English names ("Interquartile mean" instead of "Interquartilemean" etc.). Also extend the grep helper comment in SRF_Math.php with the statistical formats.

Fixes #1043


i18n/en.json
```diff
@@ -376,19 +376,19 @@
...
+	"srf_printername_mode": "Mode",
+	"srf_printername_quartillower.exc": "Lower quartile (exclusive)",
+	"srf_printername_quartillower": "Lower quartile",
+	"srf_printername_quartilupper.exc": "Upper quartile (exclusive)",
+	"srf_printername_quartilupper": "Upper quartile",
+	"srf_printername_range": "Range",
+	"srf_printername_samplestandarddeviation": "Sample standard deviation",
+	"srf_printername_samplevariance": "Sample variance",
+	"srf_printername_standarddeviation": "Standard deviation",
+	"srf_printername_variance": "Variance",
```
JeroenDeDauw, line 392: You changed `srf-printername-carousel` from underscore to dash, and then changed these ones from dash to underscore 🤔

kghbln: See the pull request comment. Saves us from changing all the keys on translatewiki.net

Nikerabbit:
We would prefer that content changes and key changes do not happen in the same commit. Then our automatic rename detection would take care of this.

kghbln:
Both points are addressed, and the branch is rebased onto the current master (the math printer moved to `src/Math/Math.php` in the meantime; the grep helper comment moved with it):

* @Nikerabbit The change is now split: the first commit renames the keys with byte-identical values, the second changes the English names only. Needs to be merged with a merge commit or rebase merge (not squash), so the two commits reach master separately, and the rename detection sees the value-identical renames.
* @JeroenDeDauw On the direction: the math family builds its name key dynamically with a single prefix (`srf_printername_` . $format) shared with the six long-established underscored keys (max, min, sum, product, average, median), so hyphenating the statistical keys would either split the family across two styles in the code or rename the established keys along with their translations. Carousel's code already requests the modern hyphenated key, so there the i18n side moves instead. Renaming each side to what the code requests keeps translatewiki churn to a minimum.

<sub>AI-drafted — Claude Code, `Fable 5 (max)`; restructure and rebase by Claude Code on @kghbln's ask.</sub>

kghbln:
Please merge with a merge commit or rebase the merge (not squash), so the two commits reach master separately, and the rename detection sees the value-identical renames.

PR #1050, opened 2026-06-19
Fix tagcloud

### LINKED ISSUE #1049: Tagcloud does not include text items
jQuery.tagcanvas only recognizes link items

```
TCproto.GetTags = function() {
  var el = this.SourceElements(), etl, tl = [], i, j;
  for(i = 0; i < el.length; ++i) {
    etl = el[i].getElementsByTagName('a');
    for(j = 0; j < etl.length; ++j) {
      tl.push(etl[j]);
    }
  }
  return tl;
};
```
(resources/jquery/jquery.tagcanvas.js)

therefore text items from TagCloud -> getTags aren't recognized

thomas-topway-it:
- fixes https://github.com/SemanticMediaWiki/SemanticResultFormats/issues/1049


formats/tagcloud/resources/ext.srf.formats.tagcloud.js
```diff
@@ -89,6 +89,17 @@
...
+			// wrap text items with a fake link to ensure
+			// JQuey's TagCloud list them
+			var tagsElement = container.find( '#' + tagsID );
+			tagsElement.find( 'li' ).each( function() {
+				var $li = $( this );
+				if ( $li.find( 'a' ).length === 0 ) {
+					var text = $li.text();
+					$li.wrapInner( '<a href="javascript:void(0)"></a>' );
+				}
+			});
```
gesinn-it-gea, line 100: **`javascript:void(0)` may break on wikis with CSP enabled**

`href="javascript:void(0)"` is treated as an inline script by browsers and is blocked by any `script-src` CSP that doesn't explicitly allow `unsafe-inline`. MediaWiki supports CSP via `$wgCSPHeader` (since MW 1.32), so this will silently prevent the injected `<a>` elements from being usable on wikis with a strict CSP — potentially causing a console error or the tag not rendering.

TagCanvas only requires an `<a>` element to be present; the `href` value itself doesn't affect sphere display. Use `href=""` instead:

```js
$li.wrapInner( '<a href=""></a>' );
```

formats/tagcloud/resources/ext.srf.formats.tagcloud.js
```diff
@@ -89,6 +89,17 @@
...
 
+			// wrap text items with a fake link to ensure
+			// JQuey's TagCloud list them
+			var tagsElement = container.find( '#' + tagsID );
+			tagsElement.find( 'li' ).each( function() {
+				var $li = $( this );
+				if ( $li.find( 'a' ).length === 0 ) {
+					var text = $li.text();
+					$li.wrapInner( '<a href="javascript:void(0)"></a>' );
+				}
```
gesinn-it-gea, line 99: **Dead variable — `text` is assigned but never used**

`var text = $li.text()` captures the element's text content, but the value is never passed to `wrapInner` or used anywhere else in the block. `wrapInner` wraps the existing child nodes in place and doesn't need it. The line can be removed.

formats/tagcloud/resources/ext.srf.formats.tagcloud.js
```diff
@@ -89,6 +89,17 @@
 
 			context.css( { 'width': data.width, 'height': data.height } );
 
+			// wrap text items with a fake link to ensure
+			// JQuey's TagCloud list them
```
gesinn-it-gea: Typo in comment: `JQuey` → `jQuery`

formats/tagcloud/resources/ext.srf.formats.tagcloud.js
```diff
@@ -89,6 +89,17 @@
...
+			var tagsElement = container.find( '#' + tagsID );
+			tagsElement.find( 'li' ).each( function() {
+				var $li = $( this );
+				if ( $li.find( 'a' ).length === 0 ) {
+					var text = $li.text();
+					$li.wrapInner( '<a href="javascript:void(0)"></a>' );
```
gesinn-it-gea: **Missing QUnit test coverage for the new wrapping logic**

The existing test in `tests/qunit/formats/ext.srf.formats.tagcloud.test.js` only uses `<li><a href="...">` elements in its fixture — the new code path (`$li.find('a').length === 0`) is never exercised. Please add a test case covering:

1. A `<li>` without an `<a>` gets wrapped after `sphere()` runs (the main fix)
2. A `<li>` that already has an `<a>` is not double-wrapped (the guard condition)

Example fixture extension:

```js
'<li>TextOnly</li>'
```

Example assertions:

```js
assert.equal( context.find( 'li:not(:has(a))' ).length, 0, 'all li elements have an anchor after sphere()' );
assert.equal( context.find( 'li:has(a > a)' ).length, 0, 'no li is double-wrapped' );
```

thomas-topway-it: done

thomas-topway-it:
@gesinn-it-gea done,  and thanks for the suggestions. I need to test the tests locally, I will be back in few hours

thomas-topway-it:
@gesinn-it-gea I have tested the tests locally and fixed an old notice 

> Tagcanvas dies during testing for some reasons

 so now the QUnit tests via `Special:JavaScriptTest` work

PR #1117, opened 2026-07-09
Enable incoming format for CI/test environments

D-Groenewegen:
Kept banging my head against a wall not understanding why integration tests consistently failed for `incoming-01.json`, even after I explicitly enabled it in `$srfgFormats`. Until now, luckily. In order for result formats to be tested that are not enabled by default, they must be enabled in the `__set_extension__` file.

Note. Should the `/tests` folder contain a README file to inform testers of this?

### DIFF
--- a/__setup_extension__
+++ b/__setup_extension__
@@ -1,8 +1,9 @@
-// Enable googlebar and googlepie formats for CI/test environments.
-// These are disabled by default in DefaultSettings.php because they contact
-// an external URL (chart.apis.google.com) to render charts. They are safe to
+// Enable googlebar, googlepie and incoming formats for CI/test environments.
+// googlebar and googlepie are disabled by default in DefaultSettings.php because they
+// contact an external URL (chart.apis.google.com) to render charts. They are safe to
 // enable here since integration tests only assert against the generated HTML,
 // not against an actual HTTP response from Google.
+// incoming is disabled by default because of its potential impact on performance.
 //
 // We register the formats directly into $smwgResultFormats and $wgAutoloadClasses
 // to bypass the srfgFormats → onExtensionFunction timing dependency.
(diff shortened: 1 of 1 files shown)


__setup_extension__
```diff
@@ -1,15 +1,18 @@
...
+// contact an external URL (chart.apis.google.com) to render charts. They are safe to
 // enable here since integration tests only assert against the generated HTML,
 // not against an actual HTTP response from Google.
+// incoming is disabled by default because of its potential impact on performance.
 //
 // We register the formats directly into $smwgResultFormats and $wgAutoloadClasses
 // to bypass the srfgFormats → onExtensionFunction timing dependency.
 $wgExtensionFunctions[] = static function () {
 	$dir = __DIR__ . '/extensions/SemanticResultFormats/formats/googlecharts/';
 	$GLOBALS['wgAutoloadClasses']['SRFGoogleBar'] = $dir . 'SRF_GoogleBar.php';
```
gesinn-it-gea: Small suggestion: `$dir` here only points at `formats/googlecharts/`, which is why the new `incoming` entry above has to spell out the full path again instead of reusing `$dir`. Making `$dir` generic (pointing at `formats/`) and appending the subdirectory per line would keep the mechanism consistent for all formats:

```php
$wgExtensionFunctions[] = static function () {
	$dir = __DIR__ . '/extensions/SemanticResultFormats/formats/';
	$GLOBALS['wgAutoloadClasses']['SRFGoogleBar'] = $dir . 'googlecharts/SRF_GoogleBar.php';
	$GLOBALS['wgAutoloadClasses']['SRFGooglePie'] = $dir . 'googlecharts/SRF_GooglePie.php';
	$GLOBALS['wgAutoloadClasses']['SRFIncoming'] = $dir . 'incoming/SRF_Incoming.php';
	$GLOBALS['smwgResultFormats']['googlebar'] = 'SRFGoogleBar';
	$GLOBALS['smwgResultFormats']['googlepie'] = 'SRFGooglePie';
	$GLOBALS['smwgResultFormats']['incoming'] = 'SRFIncoming';
};
```

PR #1125, opened 2026-07-19
fix(graph): guard against null node labels and skip-eligible page-type values

### LINKED ISSUE #1096: [graph]: nodes without labels cause deprecation warning for htmlspecialchars()
This came up during an integration test with MW 1.45, SMW 7.0.0 and PHP 8.3.

```php
Deprecated: htmlspecialchars(): Passing null to parameter #1 ($string) of type string is deprecated in /var/www/html/extensions/SemanticResultFormats/src/Graph/GraphFormatter.php on line 95
```

Since PHP 8.1, `htmlspecialchars()` strictly accepts strings only, causing a deprecation notice which in this case reveals that some nodes come without a label (`$node->getLabel()`).

gesinn-it-gea:
## Summary
- Fixes the `htmlspecialchars(): Passing null to parameter #1` deprecation notice (PHP 8.1+) in `GraphFormatter::buildGraph()` when a `GraphNode` has no label — falls back to an empty string instead.
- Fixes a related logic gap in `GraphPrinter::processResultRow()`: the node-creation check inside the `showAsEdge` branch omitted the `!$skipNode` guard applied to the first node-creation check, so a page-type value that should have been skipped (e.g. a 2nd+ page-type printout beyond the first) could still become the node.
- Added regression tests for both fixes.

## Test plan
- [x] `make dev-test` (lint → PHPCS → Phan → PHPUnit → JS tests) passes with exit code 0
- [x] Full PHPUnit suite: 457 tests, 0 failures (455 pre-existing + 2 new regression tests)
- [x] Verified the `htmlspecialchars(): Passing null...` deprecation notice from #1096 no longer appears in test output after the fix
- [x] New `testBuildGraphHandlesNodeWithoutLabelAndWithoutFields` reproduces the original crash scenario
- [x] New `testProcessResultRowDoesNotCreateNodeFromSkippedPageTypeValue` reproduces the skipNode gap

Fixes #1096

### DIFF
--- a/RELEASE-NOTES.md
+++ b/RELEASE-NOTES.md
@@ -70,6 +70,7 @@ Released on TBD.
 
 ### Bug Fixes
 
+* Fixed `graph` format: `htmlspecialchars()` deprecation notice (PHP 8.1+) when a node has no label, and a missing `skipNode` guard in `GraphPrinter::processResultRow()` that could create a node from a page-type value that should have been skipped ([1096](https://github.com/SemanticMediaWiki/SemanticResultFormats/issues/1096)) (by [gesinn.it](https://gesinn.it))
 * Fixed `graph` format crash when more than 14 distinct edge predicates are used — the color palette now cycles instead of throwing an undefined array key error (by [gesinn.it](https://gesinn.it))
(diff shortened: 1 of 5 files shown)


gesinn-it-gea:
@D-Groenewegen would you like to do a short review to confirm if this works for you

D-Groenewegen:
@gesinn-it-gea - I don't have a real-world use case at hand that I can try out. The error message was thrown by one of the CI tests for an unrelated issue on here. That said, the automated tests above all check out fine for syntax errors and everything must have passed your new regression test.

I did create a new test environment with graphviz and all, as I intended to do regardless, so if there is any concrete example you would like me to test, just let me know.

PR #1127, opened 2026-07-19
test(graph): cover processResultRow() field/edge branching and fix skipNode gap

### LINKED ISSUE #1124: [graph]: GraphPrinter core logic (processResultRow) has only 19.6% test coverage
Found during a code review of the `graph` result format (2026-07-19).

## Coverage measurement
Measured with the CI's `coverage: true` matrix entry (MW 1.43, PHP 8.3, SMW 7.0.0, MM 6.0.2, PF 5.9, SFS 4.0.0-beta — matching `.github/workflows/ci.yml`), via:
```
make install MW_VERSION=1.43 PHP_VERSION=8.3 SMW_VERSION=7.0.0 MM_VERSION=6.0.2 PF_VERSION=5.9 SFS_VERSION=4.0.0-beta
make bash
> XDEBUG_MODE=coverage composer phpunit -- --coverage-clover /tmp/cov.xml --filter Graph
```

Results for `src/Graph/`:

| Class | Line coverage | Method coverage |
|---|---|---|
| `GraphFormatter` | 98.3% | 6/6 |
| `GraphNode` | 100% | 8/8 |
| `GraphOptions` | 100% | 16/16 |
| **`GraphPrinter`** | **19.6%** | **2/7** |

`GraphPrinter::handleParameters()`, `getParamDefinitions()`, and most of `processResultRow()` — the method that turns SMW query rows into `GraphNode`s — are effectively untested. Only two narrow regression tests exist (`GraphPrinterTest::testProcessResultRowUsesWikiValueForNonPageType` / `...UsesDisplayTitleForPageType`, covering issue #988).

## Why this matters
Several logic branches in `processResultRow()` that were flagged during the same review (see #1096 comments) are entirely unexercised by the test suite, e.g.:
- The `graphfieldspages=true` field/edge branching (`$includeAsEdge`, `$includeAsField`, `$pageTypeSeen` counting).
- The `elseif ( $showAsEdge )` node-creation branch and its interaction with `$skipNode`.
- The case where `$node` stays `null` for an entire result row.

Any future change to this method could silently regress without the test suite catching it.

## Suggested fix
Add PHPUnit tests for `processResultRow()` covering:
- `graphfieldspages=true` with 0, 1, 2, and 3+ page-type printouts per row.
- Rows where every value has `getProperty()` truthy (node never created).
- The `showGraphFields`/`showGraphFieldsPages` field-vs-edge branching combinations.
- `handleParameters()` and `getParamDefinitions()` basic sanity (param keys/defaults).

Also worth a related test-quality note: `tests/phpunit/Unit/Graph/GraphFormatterTest.php` passes `'graphfieldspages' => 'no'` as a raw string directly to `GraphOptions`, bypassing SMW's `ParamProcessor` boolean normalization. Since PHP's `(bool)'no'` evaluates to `true`, these "no" test cases actually exercise the `true` code path, giving a false sense of coverage for the `false` branch.

gesinn-it-gea:
## Summary
- Adds PHPUnit coverage for `GraphPrinter::processResultRow()`'s `graphfieldspages`/`graphfields` field-vs-edge branching (0/1/2/3+ page-type printouts per row), the silently-dropped-row case (node never created), and basic `getParamDefinitions()`/`handleParameters()` sanity — closing the coverage gaps listed in #1124.
- One new test caught a real bug: the `elseif ( $showAsEdge )` node-creation branch was missing the `!$skipNode` guard already applied to the primary node-creation check, so a 3rd+ page-type value without a property could still become the node despite `$skipNode` being true (matches the gap noted in #1096's comments). Fixed in `GraphPrinter::processResultRow()`.
- Fixes `GraphFormatterTest` cases that passed `'graphfieldspages' => 'no'`/`'yes'` as raw strings instead of booleans — `(bool)'no'` is `true` in PHP, so the `'no'` cases were silently exercising the `true` branch, per the false-negative note in #1124.

Closes #1124

## Test plan
- [x] `make dev-test` (lint, PHPCS, Phan, PHPUnit) — all green, PHPUnit `462 tests, 1192 assertions`
- [x] JS QUnit suite — `198/198` passing
- [x] New tests fail against pre-fix production code (confirmed the `$skipNode` bug), pass after the one-line fix


alistair3149:
## Review

I verified the PR's central claim before anything else, and it holds: reverting the one-line `!$skipNode` guard makes exactly one test fail — the new `testProcessResultRowDoesNotCreateNodeFromSkippedPageTypeViaEdgeBranch`, which reports a spurious `GraphNode(Page3)`. So this is a genuine regression test for a genuine gap. Full Graph suite is green (`OK, 64 tests, 98 assertions`) and `composer analyze` is clean with no warnings.

The test additions are good, and the `'graphfieldspages' => 'no'` → `false` cleanup in `GraphFormatterTest` is a real correctness fix for the fixtures.

Three things below, one of which I think should be settled before this merges.

---

### 1. The guard change is a behaviour regression for `?=` queries

I originally assumed reaching this branch required `mainlabel=-`. That is not true, and correcting it changes the picture.

`QueryProcessor::addThisPrintout()` bails out when a PRINT_THIS request is already registered:

```php
// If THIS is already registered, bail-out!
foreach ( $printRequests as $printRequest ) {
    if ( $printRequest->isMode( PrintRequest::PRINT_THIS ) ) {
        return;
    }
}
```

So an explicit `?=` keeps the column position the user wrote it in, with `mainlabel` uninvolved. And `?=` parses to PRINT_THIS (`Deserializer.php:57-58`), whose `getTypeID()` returns `_wpg` (`PrintRequest.php:236-238`), with `getProperty()` null (`ResultArray.php:209, 239-240` — `$diProperty` is only set for PRINT_PROP and PRINT_CHAIN).

For `{{#ask: [[Category:City]] |?Located in |?=Page |format=graph}}`:

| column | state | before | after |
|---|---|---|---|
| `?Located in` | `_wpg`, has property, `pageTypeSeen=1` | no node | no node |
| `?=` | PRINT_THIS, `_wpg`, **no** property, `pageTypeSeen=2` → `$skipNode` true | node = **the subject page** | nothing |

Before this PR that row rendered an isolated node correctly named after the subject (the edge was already lost). After it, because `if ( $node )` at line 291 gates the parents, the fields *and* `$this->nodes[]`, the whole row silently produces nothing.

Both behaviours are wrong, but this turns "right node, missing edge" into "row absent with no warning", which is harder to notice on a wiki.

The mirror case is also worth noting: when the property-less page-type column is **first**, `pageTypeSeen == 1`, `$skipNode` is false, and a junk category-named node is still created. The guard only covers position 2 and later.

Both directions point at the same root cause — node identity is inferred from *column position* rather than from the query subject. Deciding that seems like the right precondition for this fix.

### 2. After the fix, lines 245-247 are unreachable

```php
} elseif ( $showAsEdge ) {
    if ( !$node && !$hasProperty && !$skipNode ) {   // <- can never be true
```

Within `processResultRow()`, `$node` is assigned only at 191/230/246, `$hasProperty` only at 215, `$skipNode` only at 226. Line 244 is the `elseif` of the block starting at 234, so when line 245 is reached nothing at all has executed since 232. The condition is therefore evaluated on identical operands to line 229: either 229 created the node (so `!$node` is false), or 229 was false for exactly the same reason. Either way 245 is false.

This is copy-paste residue from the hoist in 850d92ba (#958), where the inner block was the only node-creation path. Collapsing it is a strict no-op:

```php
} elseif ( $showAsEdge ) {
    if ( $node && $objectText !== $node->getId() ) {
        $parents[] = [
            'predicate' => $label,
            'object' => $objectText,
        ];
    }
    continue;
}
```

I applied that locally and the full Graph suite stays green with `analyze` clean. Worth doing — but note it removes this PR's entire production diff, so the commit message and release note would need rewriting from "was missing the `!$skipNode` guard" to "remove duplicated node creation left over from #958". Which is really the accurate description either way: as written, the patch deletes a node-creation path rather than repairing one.

### 3. `3rd+` should be `2nd or later`

`$skipNode = $isPageType && $pageTypeSeen > 1`, so it becomes true on the **second** page-type printout, not the third. The current wording appears in RELEASE-NOTES.md, the commit body, and the docblock at `GraphPrinterTest.php:299-304`.

Confirmed empirically — a two-column row (page-type printout 1 whose value has a property, printout 2 whose value does not) creates a spurious node against the pre-fix code and none after:

```
✘ Temp two column skip node
  Failed asserting that two arrays are identical.
  +    0 => SRF\Graph\GraphNode Object ( 'id' => 'Page2' ... )
```

It matters because a reader checking their wiki for queries with three or more page-type printouts would wrongly conclude they were never affected. Two is enough.

The regression test itself uses three columns where two suffice, so tightening it to the minimal case would pin the actual boundary.

(The `3+` wording elsewhere is correct and should be left alone — the data-provider key and the Tests release-note entry describe the separate `graphfieldspages` field branch, whose threshold really is `$pageTypeSeen > 2`.)

Also worth stating in the release note: suppressing node creation drops the row's edges and fields too, not just the node.

---

### Minor

- Two surviving mutants in the new suite: `$pageTypeSeen > 2` → `>= 2` at line 261, and moving the `$pageTypeSeen++` at line 204 inside the inner `while`. Both leave the whole suite green. The first is closable with one provider row — `[ [ '_wpg', false ], [ '_wpg', true ] ]` expecting a parent and no field.
- The `(graphfields=false, graphfieldspages=true)` combination is untested, though #1124 asked for the branching combinations. It is reachable via `|graphfieldspages=yes` alone.
- Multi-value printouts are never exercised — every `makeResultArray()` call site passes a single value, so the inner `getNextDataValue()` loop only ever runs once.
- `makePrintRequest()` gained `$label` and `$isChain` parameters that no call site passes, so the PRINT_CHAIN branches stay uncovered and the `isMode` callback rewrite is inert.

### Unrelated, pre-existing

`srf-paramdesc-graphfieldspages` is referenced at `GraphPrinter.php:416` but is absent from `i18n/` entirely — the parameter has no description anywhere. Probably worth its own issue.

gesinn-it-gea:
Thank you for this review — genuinely one of the most thorough we've received. We independently re-derived each of your three points before acting on them (formally simulating the control flow rather than just trusting the description), and all three held up. Two new commits address them.

**One heads-up first:** `#1125` (merged to `master` shortly before this review landed) independently made the same one-line `!$skipNode` guard fix that this PR originally had — before this review caught that it was patching a duplicate rather than removing it. That created a rebase conflict, now resolved: this branch has been rebased onto current `master`, and the commits below supersede `#1125`'s guard with the corrected fix. `#1125`'s unrelated `htmlspecialchars(null)` fix is untouched.

**Point 2 (dead code) — [b10532c](https://github.com/SemanticMediaWiki/SemanticResultFormats/pull/1127/commits/b10532c7)**

Confirmed exhaustively: with `$node`/`$hasProperty`/`$skipNode` unchanged between line 229 and the `elseif ( $showAsEdge )` duplicate, and the `if`/`elseif` structure ensuring the duplicate is only reached when the `showGraphFieldsPages` branch didn't run, the duplicate's condition evaluates identically to line 229's on every one of the 16 reachable input combinations we checked. Removed it rather than patching it — commit message and RELEASE-NOTES rewritten accordingly ("remove duplicated node-creation check" instead of "add missing guard").

**Point 3 (2nd vs 3rd) — same commit**

Correct, `$skipNode = $isPageType && $pageTypeSeen > 1` triggers from the 2nd page-type printout, not the 3rd. Fixed in the regression test (now minimal at 2 columns instead of 3), its docblock, and RELEASE-NOTES. Left the separate `graphfieldspages` field threshold (`$pageTypeSeen > 2`, genuinely 3rd+) untouched, as you noted.

**Point 1 (`?=` regression) — [3d4a406](https://github.com/SemanticMediaWiki/SemanticResultFormats/pull/1127/commits/3d4a406a)**

You were right that this was the real bug hiding under the original one-line fix, and that it deserved more than a footnote — this was the actual defect #1124 was chasing. Root cause as you diagnosed: node identity was inferred from column position (`$pageTypeSeen`) instead of the query subject.

Fix: `rowHasThisPrintout()` scans the row up front for a `PRINT_THIS` printout. If one exists, it's always the node candidate regardless of position, and every other page-type printout is skipped as a candidate. Falls back to the old position-based behavior when there's no `PRINT_THIS` printout at all, so rows without an explicit/implicit subject column are unaffected.

Verified against both scenarios from your comment (independently simulated in an isolated PHP harness before touching the real code):

| scenario | on `master` (before #1127) | after the guard fix, before the `?=` fix | after the `?=` fix (3d4a406) |
|---|---|---|---|
| `?Located in` (prop) then `?=` (no prop) | node = subject, edge lost | row silently dropped | node = subject, edge preserved |
| mirror case: property-less category column before `?=` | bogus node named after the category value | (unchanged) | node = subject |
| no `PRINT_THIS` in the row at all | position-based | unaffected | unaffected |

Two new regression tests (`testProcessResultRowUsesThisPrintoutAsNodeRegardlessOfPosition`, `testProcessResultRowPrefersThisPrintoutOverEarlierPageTypeColumn`) cover both directions.

Also applied your two smaller suggestions from the "Minor" section: added the `[ [ '_wpg', false ], [ '_wpg', true ] ]` provider row (closes the `> 2` → `>= 2` mutant at the field threshold), and a test for the untested `graphfields=false`/`graphfieldspages=true` combination.

Left as-is / out of scope for this PR, flagging for follow-up rather than silently dropping:
- PRINT_CHAIN branches remain uncovered (`makePrintRequest()`'s `$isChain` param is still unused by any call site) — a real gap, but a separate coverage task.
- Multi-value printouts per column are still untested (every `makeResultArray()` call site passes a single value).
- The missing `srf-paramdesc-graphfieldspages` i18n message is unrelated to this PR's scope.

Full suite green throughout, including after the rebase onto `master` (`#1125` + others): `make dev-test` (lint, phpcs, phan, phpunit) and npm/QUnit all pass — `506 tests, 1257 assertions` PHPUnit, `213/213` QUnit.

alistair3149 (approved):

gesinn-it-gea:
**Follow-up on the `?=` fix above:** writing an integration test that asserts against the actual DOT source (`GraphFormatter::getGraph()`, i.e. exactly what gets handed to `dot`/the `<graphviz>` tag) — rather than only the in-memory `GraphNode` from `processResultRow()` — immediately caught a second, related bug the node-only unit tests had missed: the edge from `?Located in` to the `?=` subject was silently lost whenever the subject column came after the property-bearing column. `processResultRow()` built edges inline while iterating the row, so the node (only known once the later `?=` column was reached) was still `null` while the earlier column's edge was being decided.

Fixed in [25a2026](https://github.com/SemanticMediaWiki/SemanticResultFormats/pull/1127/commits/25a2026d) by splitting `processResultRow()` into three explicit steps — `readRowValues()` (read every value out of the row once, since `ResultArray`'s iterator is single-pass), `determineNode()` (pick the node from the full classified row), `collectEdgesAndFields()` (build edges/fields against the already-known node) — instead of patching the inline version further. Two new tests assert against the real DOT output for both PRINT_THIS scenarios.

Given how effective that turned out to be, DOT-source assertions (not just in-memory node checks) feel like the right default for future `graph`-format regression tests in this file.

gesinn-it-gea:
**One more follow-up, building on the DOT-source methodology above:** since asserting against the actual DOT source caught a real bug, I applied it systematically across `GraphPrinter::getParamDefinitions()`'s full parameter list, not just the PRINT_THIS scenarios — 14 new tests in [a495586](https://github.com/SemanticMediaWiki/SemanticResultFormats/pull/1127/commits/a4955864), covering `graphname` (sanitization), `graphfontsize`, `graphsize`, `nodeshape`, `arrowdirection` (all four `rankdir` values), `graphlink`, `nodelabel` (`displaytitle` word-wrap), `graphlabel`/`graphcolor`/`arrowhead` (edge attributes, including the "only rendered together" guard), `relation` (edge direction), and `graphlegend` (only with `graphcolor`). Each builds the real `GraphPrinter` → `GraphFormatter` pipeline and asserts against the same DOT string that gets handed to the `<graphviz>` tag / the `dot` binary.

Two mistakes surfaced along the way, both in the new tests themselves rather than in `GraphPrinter`/`GraphFormatter`:
- `assertStringNotContainsString('size=', $dot)` false-failed because `'size='` is a substring of the always-present `fontsize=` — narrowed to `'size="'`.
- Reusing the same row array across two `buildDot()` calls in one test silently returned no node on the second call, because `ResultArray`'s value iterator (mocked here) is single-pass — switched to a factory closure that builds fresh mocks per call.

No further bugs in production code from this pass — the remaining parameters behaved as their implementation suggests. `wordwraplimit` and `graphfields`/`graphfieldspages` are already exercised by the `nodelabel` test and the earlier field-branching tests respectively, so they weren't duplicated.

`buildFormatter()`/`buildDot()` test helpers now back both this and the earlier DOT-source tests, so future `graph`-format regression tests in this file have a ready-made way to assert against real DOT output instead of just the in-memory `GraphNode`.

PR #1128, opened 2026-07-19
ci(mediawiki): install Diagrams extension to test the graph format's Diagrams-specific rendering path

### LINKED ISSUE #846: Graph format shows "<br />" when specifying nodelabel
### Setup

- MW version: 1.39.8
- DB (MySQL etc.): 10.5.23-MariaDB-0+deb11u1
- PHP version: 7.4.33 (apache2handler)
- SMW version: 4.2.0
- SRF version: 4.2.1
- Browsers and versions (if applicable): n/a

### Issue

Graphs with overridden nodelabel display `<br />` instead of line breaks in the labels.

**Steps to reproduce**

Create a graph and specify `|nodelabel=displaytitle`.

Example on SMW sandbox: https://sandbox.semantic-mediawiki.net/wiki/Mise_en_place_d%27un_wiki_(small)

**Expected result:**
The labels in the graph should contain line breaks.

**Observed result:**
The labels in the graph contain `<br />` instead of line breaks.

Screenshot_20240801_202649

gesinn-it-gea:
## Summary
- The `graph` format's `GraphFormatter`/`GraphPrinter` branch on `ExtensionRegistry::isLoaded('Diagrams')`, switching between an HTML `<br />` line separator and a plain `PHP_EOL` one for word-wrapped node labels. CI never had `Diagrams` installed, so that entire branch — including the bug reported in #846 — went untested.
- Bumped the `build/` docker-compose-ci submodule (4 commits, all additive/defensive) to pick up its `extensions.local.json` support for non-bundled custom extensions, plus SMW 7 compatibility (which incidentally silences a `Deprecated: Use of enableSemantics` notice that showed up on every test run).
- Added `extensions.local.json` to install `Diagrams` ([samwilson/diagrams-extension](https://github.com/samwilson/diagrams-extension), tag `1.0.0`) and added `graphviz` to `OS_PACKAGES` in the `Makefile` (provides the `dot` binary Diagrams shells out to for local rendering — confirmed no network access is involved: `DiagramsServiceUrl` stays unset, so Diagrams uses `renderLocally()`, which explicitly calls `disableNetwork()` on the shell command).
- Fixed `GraphFormatterTest::testGetWordWrappedText`, which had been implicitly assuming Diagrams is never loaded (its expectations were hardcoded to the `PHP_EOL` separator) — it now derives the expected separator from `ExtensionRegistry::isLoaded('Diagrams')`, same as the production code.

## Why
Preparing to fix #846 (graph format shows literal `<br />` instead of a line break when `nodelabel=displaytitle` produces a wrapped label under Diagrams). Verifying that fix properly requires exercising the actual `Diagrams`-loaded code path in CI, not just the `PHP_EOL` fallback that happened to be the only one ever tested.

## Test plan
- [x] `make dev-test` (lint → PHPCS → Phan → PHPUnit → JS tests) passes with exit code 0, 453 PHP tests green
- [x] Verified `dot -V` resolves inside the container after the `graphviz` OS package is installed
- [x] Verified the `Diagrams` extension is actually registered (`wfLoadExtension( "Diagrams" )` present in `__setup_extension__`) and `ExtensionRegistry::isLoaded('Diagrams')` returns `true` in a test
- [x] Verified no network calls occur: confirmed `DiagramsServiceUrl` defaults to empty (falls back to local rendering) and that `renderLocally()` calls `disableNetwork()` on its shell command
- [x] Confirmed the previously-passing `testGetWordWrappedText` failed once Diagrams was actually loaded (proving the test was environment-dependent) and passes again after the fix, under both Diagrams-loaded and Diagrams-not-loaded conditions
- [x] End-to-end reproduced #846 with the real `dot` binary: rendered SVG for a wrapped label under `Diagrams` contains the literal text `long&lt;br /&gt;label&lt;br /&gt;text&lt;br /&gt;here` instead of line breaks — confirms this setup surfaces the real bug for a follow-up fix PR


gesinn-it-gea:
Rebased this branch onto current `master` (which now includes #1125 and #1127) — no conflicts, since this PR only touches `Makefile`, `extensions.local.json`, the `build/` submodule bump, and `GraphFormatterTest`.

Rebuilding the image with `Diagrams`/`graphviz` actually installed (rather than just rebasing the source) surfaced one more environment-dependent test: `GraphPrinterTest::testNodeLabelDisplaytitleWordWrapsNodeLabelInDotSource` (added in #1127's DOT-source test pass) hardcoded `PHP_EOL` as the `nodelabel=displaytitle` word-wrap separator. With `Diagrams` loaded, `GraphFormatter` emits `<br />` instead — not a regression, but the same environment-dependent behavior `GraphFormatterTest::testGetWordWrappedText` already had to account for in this PR. Fixed in [311ab18](https://github.com/SemanticMediaWiki/SemanticResultFormats/pull/1128/commits/311ab18a) the same way: derive the expected separator from `ExtensionRegistry::isLoaded('Diagrams')` at assertion time instead of hardcoding it.

Confirms this PR's own point: exercising the real `Diagrams`-loaded path immediately finds environment assumptions baked into tests that only ever ran against the `PHP_EOL` fallback. Full suite green with `Diagrams`+`graphviz` actually installed: `524 tests, 1289 assertions` PHPUnit, `213/213` QUnit.

PR #1129, opened 2026-07-19
fix(graph): render <br />-wrapped node labels as HTML labels, not quoted strings

### LINKED ISSUE #846: Graph format shows "<br />" when specifying nodelabel

gesinn-it-gea:
## Summary
- Fixes #846: the `graph` format showed a literal `<br />` instead of a line break in node labels wrapped under the Diagrams extension (e.g. `nodelabel=displaytitle` with a long display title).
- Root cause: `GraphFormatter::buildGraph()`'s fieldless-node branch re-escaped the already-escaped, already-wrapped label with a second `htmlspecialchars()` call and wrapped it in a quoted DOT string label (`"..."`). Under Diagrams, wrapped labels are joined with a literal `<br />`; a string label renders that as literal text (`&lt;br /&gt;`), not a line break — GraphViz only interprets HTML markup inside an HTML label (`<...>`, unquoted).
- Fix: drop the redundant re-escaping, and switch to an HTML label whenever the wrapped text actually contains `<br />` — matching how the `graphfields` (on-node fields) branch already handles multi-line labels.

## Why this branch is based on #1128
Verifying this fix requires exercising the Diagrams-specific code path, which #1128 sets up (installs the `Diagrams` extension + `graphviz`'s `dot` binary in CI). This PR's diff will shrink to just the fix once #1128 is merged.

## Test plan
- [x] `make dev-test` passes with exit code 0, 455 PHP tests green (453 + 2 new)
- [x] End-to-end verified with the real `dot` binary (not just DOT-source string comparison):
  - Before the fix: rendered SVG text elements were `["long&lt;br /&gt;label&lt;br /&gt;text&lt;br /&gt;here"]` — the exact bug from the issue's screenshot.
  - After the fix: rendered SVG text elements are `["long","label","text","here"]` — four separate lines.
- [x] Added `testBuildGraphRendersLineBreaksAsHtmlLabelUnderDiagrams` — forces the Diagrams-style separator via reflection (independent of whether Diagrams happens to be loaded in the running test environment) and asserts the DOT source uses an HTML label with no re-escaped `<br />`.
- [x] Added `testBuildGraphKeepsQuotedLabelWhenNoLineBreakIsNeeded` — a short label that never gets wrapped stays a plain quoted string label, confirming no regression for the common case.
- [x] Verified the non-Diagrams (`PHP_EOL` separator) case is unaffected: still renders as a quoted string label, and `dot` still renders it as multiple lines (GraphViz tolerates raw newlines in string labels, even though `\n` is the documented escape).


gesinn-it-gea:
Rebased onto `master` (this PR's base was auto-updated after #1128 merged) — no conflicts other than two expected ones, both resolved by keeping both sides (independent additions, not competing changes):
- `RELEASE-NOTES.md`: this PR's #846 entry alongside the unrelated entries #1127/#1128 had added.
- `GraphFormatterTest.php`: this PR's two new tests alongside #1127's unrelated `testBuildGraphHandlesNodeWithoutLabelAndWithoutFields`.

One more thing surfaced once the image was actually rebuilt with `Diagrams`/`graphviz` (from #1128) and this PR's fix applied together: `GraphPrinterTest::testNodeLabelDisplaytitleWordWrapsNodeLabelInDotSource` (added in #1127's DOT-source test pass, before this fix existed) still expected the pre-fix, buggy output under Diagrams — a quoted string label with the literal, re-escaped `&lt;br /&gt;` text. Updated in [f0e3d21](https://github.com/SemanticMediaWiki/SemanticResultFormats/pull/1129/commits/f0e3d21c) to expect this PR's actual fixed behavior instead: an HTML label (`<...>`) with lines joined by a real `<br />` under Diagrams, or the previous quoted string label with `PHP_EOL`-joined lines otherwise — mirroring how this PR's own `testBuildGraphRendersLineBreaksAsHtmlLabelUnderDiagrams` already asserts it at the `GraphFormatter` level.

Full suite green with `Diagrams`+`graphviz` actually installed: `526 tests, 1292 assertions` PHPUnit, `213/213` QUnit.

PR #1133, opened 2026-07-23
[Filtered] compacted client payload, base-layer switcher, and GeoJSON overlays

kghbln:
Three changes to the filtered format:

* **Compacted client payload (v2 serialization)**: per-item printouts become a positional array of `{v, f, s}` slots (formatted/sort values omitted when identical, value-less printouts as `null`), coordinates single-sourced. Decoded `srfFilteredConfig` −53% (3.57 → 1.68 MB on a 2,575-item page), page −37%, identical behavior. Gadgets reading `srfFilteredConfig` must adapt (release note included).
* **`map view layers` parameter**: multiple base layers with a Leaflet switcher. Names are leaflet-providers strings and/or custom layers defined by the Maps extension's new `$egMapsLeafletLayerDefinitions` (Maps 13.1+, ProfessionalWiki/Maps#921); without Maps, provider names still work and unknown names are skipped without breaking the map.
* **`map view geojson` parameter**: a GeoJSON overlay from a `GeoJson:` page or URL, fetched through the Maps extension (SSRF-guarded), simplestyle styling and popups, toggleable via the layer switcher, unaffected by the client-side filtering. With Maps absent the parameter yields an i18n'd query error and the view renders normally.

All three parts were independently pr-reviewed (report-only) with findings integrated: numeric layer names keep their configured order, malformed GeoJSON degrades to no-overlay instead of a blank map, value-filter radio DOM ids are page-unique, and GeoJSON point features use the explicitly-pathed marker icon.


JeroenDeDauw (approved):
LGTM. Tested locally

PR #1136, opened 2026-07-28
fix(gallery): render the further-results link's SMW 7 marker as HTML, not escaped text

### LINKED ISSUE #1058: [gallery] Default further results label should render
## Setup

- [MediaWiki](https://www.mediawiki.org/)	[1.43.8](https://www.mediawiki.org/wiki/MediaWiki_1.43) [(82bc9a8)](https://github.com/wikimedia/mediawiki/commit/82bc9a824508fd27b4c2d2e91f76103fc9d798f3)
- [PHP](https://php.net/)	8.3.31 (apache2handler)
- [MariaDB](https://mariadb.org/)	11.8.8-MariaDB-ubu2404
- [Semantic MediaWiki](https://www.semantic-mediawiki.org/)	7.0.0
- [Semantic Result Formats](https://www.semantic-mediawiki.org/wiki/Extension:Semantic_Result_Formats)	6.0.0

## Issue
For a query like, e. g., the further results search label gets rendered raw
```
{{#ask:
 [[File:+]]
 [[Modification date::>>{{#time: Y-m-d|now - 30 days}}]]   
 |format=gallery
 |sort=Modification date
 |order=desc
 |limit=5
}}
```
`<span class="smw-localized-message" data-smw-msg="further-results">… ytterligare resultat</span>`

Image

To mitigate until the fix was authored one need to explicitly set the `searchlabel` parameter
```
{{#ask:
 [[File:+]]
 [[Modification date::>>{{#time: Y-m-d|now - 30 days}}]]   
 |format=gallery
 |sort=Modification date
 |order=desc
 |limit=5
 |searchlabel=… ytterligare resultat
}}
```

gesinn-it-gea:
## Summary
- `gallery`'s "further results" link was built in `SMW_OUTPUT_HTML` mode, which HTML-escapes the search label
- SMW 7 defaults that label to a `<span class="smw-localized-message">` marker meant to survive as unescaped markup and be resolved after the parser cache (`ParserOutputPostCacheTransform`) — escaping it broke resolution, leaving the raw marker visible on the page (#1058)
- SMW core's own `list`/`table`/`category` printers avoid this by always building the further-results link in wiki mode; `gallery` now does the same and parses the result to HTML, since its overall output is marked `isHTML`
- Confirmed live against a real SMW 7.0.0 install: before the fix, the page rendered `<span class="smw-localized-message" data-smw-msg="further-results">… weitere Ergebnisse</span>` as literal text inside the link; after the fix, it renders the resolved `… weitere Ergebnisse` text, matching `format=list`'s output for the same query
- Note: a `searchlabel` value containing markup (e.g. `<b>text</b>`) is now rendered as HTML instead of escaped text — this brings `gallery` in line with `list`/`table`/`category`, which already behave this way

Fixes #1058

## Test plan
- [x] `composer lint` / `composer phpcs` / `composer phan` — clean
- [x] `composer phpunit` — 555 tests, 1337 assertions, all passing (including a new regression test asserting the marker survives unescaped)
- [x] `npm run test` — 241 node-qunit tests passing
- [x] Manually reproduced the bug and verified the fix against a live MediaWiki 1.43 + SMW 7.0.0 + SRF install


alistair3149:
Reviewed against a live MediaWiki 1.43 + SMW 7.0.0-alpha stack, with this branch and `master` as two separate installs. The diagnosis is right and the fix works for normal page views, but the `isSpecialPage()` guard introduces a security regression that I think blocks merge.

### Confirmed working

`{{#ask: [[Has probe::yes]] |format=gallery |limit=1 }}` with three matching pages:

- `master` renders `&lt;span class=&quot;smw-localized-message&quot; data-smw-msg=&quot;further-results&quot;&gt;... further results&lt;/span&gt;` as literal text — #1058 reproduced.
- This branch renders `<a href="/index.php/Special:Ask/...">... further results</a>`. Fixed.

### Blocking: the `!isSpecialPage()` branch is unsanitized

```php
if ( !$this->isSpecialPage() ) {
    $furtherResultsLink = ...->recursiveTagParse( $furtherResultsLink );
}
$html .= $furtherResultsLink;
```

When that guard is true the link is appended as unparsed wikitext, and `getResultText()` returns it with `'isHTML' => true` — so MediaWiki splices it into the page verbatim and nothing sanitizes it. `searchlabel` is user-authored wikitext, so it reaches the browser as raw HTML.

Repro with PageForms, which ships in the CI image. Given three pages carrying `[[Has probe::yes]]`, save this as `Form:XssVerify`:

```
<noinclude>
This is the form.
</noinclude><includeonly>
{{{info|}}}
{{#ask: [[Has probe::yes]]
 |format=gallery
 |limit=1
 |searchlabel=<script>document.title='PWNED_VERIFY'</script>XSSMARK
}}
{{{standard input|save}}}
</includeonly>
```

Then request `index.php?title=Special:RunQuery/XssVerify`. On this branch the response body contains:

```
...format=gallery/searchlabel=-3Cscript-3E...|<script>document.title='PWNED_VERIFY'</script>XSSMARK]]
```

Raw `<script>` tags, inside an unrendered `[[...]]`. It executes — confirmed in a browser, `document.title` becomes `PWNED_VERIFY`. On `master` the identical form renders:

```
...">&lt;script&gt;document.title=&#39;PWNED_VERIFY&#39;&lt;/script&gt;XSSMARK</a>
```

Escaped. So this is a regression introduced here, not a pre-existing hole. `Special:FormEdit` behaves the same way, and `action=expandtemplates` returns the raw wikitext link.

That also corrects one point in the PR description: this does not bring `gallery` in line with `list`. I ran the same payload through `format=list` on the same path and it escapes on both branches — `ListResultPrinter` returns wikitext *without* `isHTML`, so the parser always sanitizes it. Gallery's `isHTML => true` return means it has no such backstop, so the skipped-parse branch is more permissive than anything `list`/`table`/`category` do.

The fix is the `RecursiveTextProcessor` SMW already injects into every printer (`QueryProcessor` calls `setRecursiveTextProcessor()` with the in-flight parser). SRF uses this pattern in `formats/datatables/DataTables.php:479-525`:

```php
if ( $this->recursiveTextProcessor === null ) {
    $this->recursiveTextProcessor = new RecursiveTextProcessor();
}

$html .= $this->recursiveTextProcessor->recursiveTagParse( $furtherResultsLink );
```

`RecursiveTextProcessor::recursiveTagParse()` guards on `$parser->getOptions() && $parser->getPage()` and falls back to a full `parse()` when no parse is in progress, so there is no unparsed branch left to leak. It also uses the parser SMW handed the printer rather than `ParserFactory::getMainInstance()`, which is a different object during a nested parse.

Worth noting so you don't over-scope the fix: `api.php?action=parse` is *not* affected, since `ApiParse` reassigns `$wgTitle` to the target page before parsing. The reachable cases are genuine Special-page renders.

### The new test passes without the fix

`testGetResultTextResolvesDeferredLocalizedSearchLabelInsteadOfEscapingIt` does not fail on pre-fix code. Applying only the test file on top of `master`, leaving `Gallery.php` at master:

```
$ git checkout fix/gallery-further-results-marker-1058 -- tests/phpunit/Unit/Formats/GalleryTest.php
$ composer phpunit -- --filter testGetResultTextResolvesDeferredLocalizedSearchLabel --testdox

 ✔ Get result text resolves deferred localized search label instead of escaping it
OK (1 test, 3 assertions)
```

`createDefaultParamMocks()` has no `searchlabel` entry, and `getQueryLink` is stubbed with `willReturn( new Infolink( true, $marker, 'Special:Ask' ) )`, which discards the caption argument — so `getSearchLabel()`, where the `htmlspecialchars()` actually happens, is never reached. On pre-fix code `Infolink::getText( SMW_OUTPUT_HTML )` then goes through `Linker::link( $title, $caption )`, whose second parameter is raw HTML, so the marker comes out unescaped and both assertions hold either way.

Passing the marker in as the `searchlabel` parameter and letting `getQueryLink` build the `Infolink` from its argument gives the test discriminating power:

```php
$queryResult->method( 'getQueryLink' )
    ->willReturnCallback( fn ( $caption ) => new Infolink( true, $caption, 'Special:Ask' ) );
```

A JSONScript case alongside `tests/phpunit/Integration/JSONScript/TestCases/gallery-0*.json` would be worth adding too — those run real `#ask` queries through the real parser, and would have caught the sanitization issue above as well.

### Minor, pre-existing

`getLink()` rather than `getFurtherResultsLink()`. The offset is not missing — SMW's `QueryResult::getQueryLink()` sets it — but it is not re-applied after `getLink()`'s `fullParams` loop, so an `#ask` with an explicit `|offset=` repeats the current window instead of advancing:

| query | gallery | list |
| --- | --- | --- |
| `limit=1` | `offset=1` | `offset=1` |
| `limit=1 offset=1` | `offset=1` | `offset=2` |

Gallery's link is also missing the `smw-gallery-furtherresults` class that `list` and `table` get. Both come free from switching to `getFurtherResultsLink()` on the line you are already touching, though a follow-up is fine.

alistair3149:
Correction to my comment above — it makes the blocking issue worse, not better.

I wrote that `api.php?action=parse` is not affected and that "the reachable cases are genuine Special-page renders". The first half holds; the second is wrong. `ApiPurge` does not reassign `$wgTitle` the way `ApiParse` does — `ApiPurge.php:114` calls `$page->updateParserCache()` with `$wgTitle` still set to `Special:Badtitle/dummy title for API calls set in api.php`. So the unparsed branch is taken *and the result is written into the ParserCache*, then served to every subsequent reader.

No extension required, and no login. Given an ordinary article containing `{{#ask: [[Has probe::yes]] |format=gallery |limit=1 }}`:

```
$ curl -s 'index.php?title=P' | grep -c '\[\[:Special:Ask'
0                      # correct: <a href="...">... further results</a>
                       # footer: Rendering was triggered because: page-view

$ curl -s -d 'action=purge&forcelinkupdate=1&titles=P' api.php     # anonymous, unauthenticated

$ curl -s 'index.php?title=P' | grep -c '\[\[:Special:Ask'
1                      # footer: Rendering was triggered because: api-purge
                       # <p>[[:Special:Ask/...|... further results]]
```

The same sequence on `master` returns `0` both times, so this is introduced by the new guard.

With a `searchlabel` payload the two issues compose, and PageForms drops out of the picture — an ordinary content page is enough:

```
|searchlabel=<script>document.title='PWNED_PURGE'</script>PURGEMARK

view 1                     raw <script> count 0    (escaped)
anonymous api.php purge
view 2 and onward          raw <script> count 1    <script>document.title='PWNED_PURGE'</script>
```

That is stored XSS on a normal article, persisted in the ParserCache, triggerable by an unauthenticated request.

The `RecursiveTextProcessor` change suggested above still resolves it, since it leaves no unparsed branch — but the severity is higher than I first described, and the `!$this->isSpecialPage()` condition should not survive in any form.

gesinn-it-gea:
@alistair3149 Thank you for the thorough review and for actually reproducing this against a live install — that XSS would have been an embarrassing one to ship, and your `ApiPurge` follow-up in particular (unauthenticated, persisted in the parser cache) is a much more serious finding than I'd realized when I opened this PR. Really appreciate the depth here.

Pushed a fix addressing all three points:

**The XSS regression:** Replaced the `!isSpecialPage()` guard around `recursiveTagParse()` with SMW's `RecursiveTextProcessor` — the same pattern `DataTables.php` already uses in this codebase. Its `recursiveTagParse()` unconditionally resolves to HTML, falling back to a full `parse()` when none is in progress, so there's no branch left where an unresolved value (including a user-supplied `searchlabel`) can reach the page unparsed. I reproduced both your `PageForms`/`Special:RunQuery` case and the anonymous `ApiPurge` case against a live SMW 7.0.0 install before and after — confirmed the payload now comes out escaped in both.

**The non-discriminating test:** You were right that `testGetResultTextResolvesDeferredLocalizedSearchLabelInsteadOfEscapingIt` passed against unfixed code — the `getQueryLink()` mock returned a fixed `Infolink` regardless of the caption it was given, so `getSearchLabel()`'s `htmlspecialchars()` call was never actually exercised. Replaced it with a mock that threads the caption through via `willReturnCallback()`, and added a new test (`testGetResultTextSanitizesSearchLabelWhenNoParseIsInProgress`) that swaps in a fresh `Parser` via `redefineService()` — the same pattern `TreeTest::replaceParser()` uses elsewhere in this suite — to reliably force the "no parse in progress" condition regardless of test execution order. Confirmed it fails against the previous fix and passes now.

**JSONScript coverage:** Added `gallery-04.json`, exercising the fix through both a real `#ask` page render and a direct `Special:Ask` call, with a `searchlabel` payload asserting no raw `<script>` or unresolved `[[Special:Ask...` wikitext reaches the output.

The `getLink()` vs. `getFurtherResultsLink()` offset/CSS-class point is still open — agreed a follow-up is fine there, will track separately rather than scope-creep this PR further.

Full CI is green across the matrix (https://github.com/SemanticMediaWiki/SemanticResultFormats/actions/runs/30400308055). Let me know if anything else needs another look.

PR #1142, opened 2026-08-03
fix(graph): always wrap field values with <br />, not lineSeparator (#816)

### LINKED ISSUE #816: Bug [graph]: fix word wrapping in on-edge fields
In graph format, non-page property values that are displayed on nodes (with `graphfields = yes`) should be word-wrapped with `<br />`, not newline, since they are displayed within `<td>` HTML tags.

gesinn-it-gea:
## Summary

- Fixes #816: in the `graph` format, field values shown on nodes (`graphfields=yes`) are always rendered inside a `<td>` of a DOT HTML label — where only a literal `<br />` is interpreted as a line break by `dot`. `GraphFormatter` used to pick the word-wrap separator based on `ExtensionRegistry::isLoaded('Diagrams')` (`<br />` if loaded, a plain newline otherwise), an assumption that was never actually true for this code path: a plain newline inside an HTML label's `<td>` is silently swallowed, running wrapped words together.
- Field values and the fields-table header now always wrap with `<br />`, independent of whether Diagrams is loaded. The Diagrams-dependent separator is kept only for the one remaining case where it's actually needed: a field-less node's own label, which can render as either a quoted DOT string or an HTML label depending on content (see #846).
- While refactoring `buildGraph()`'s control flow, caught (via review) and fixed a regression the refactor itself introduced: a node's caption must be used as the fields-table header whenever set (`GraphNode::setLabel()` is called unconditionally by `GraphPrinter::determineNode()`), regardless of the `nodelabel` option — the initial refactor had accidentally gated that fallback on `nodelabel=displaytitle`.
- Added unit regression tests for the field-value, `_wpg`-field-value, and fields-table-header wrap paths (closing coverage gaps found via a systematic branch-coverage audit), plus general `graphlink=false` coverage that was previously entirely untested in this file.
- Added `tests/phpunit/Integration/Graph/GraphvizRenderingTest.php`: integration tests that render through the real `dot` binary (no mocking), including a full end-to-end smoke test of External Data's `<graphviz>` tag-emulation path — real config, real `EDConnectorBase::loadConfig()`, real `Parser::setHook()`/`Parser::parse()` cycle, real shell-out to `dot`.
- Added `ExternalData` to the CI container's `extensions.local.json` (download-only, not `wfLoadExtension`'d by default, to avoid colliding with the already-loaded `Diagrams` extension's competing `<graphviz>` tag handler — both extensions call `Parser::setHook('graphviz', ...)` and MediaWiki keeps only the last registration).

## Why

Reported in #816: non-page property values displayed on graph nodes should wrap with `<br />`, not a plain newline, since they're rendered inside `<td>` HTML tags. Verifying the fix properly required establishing, with real `dot` rendering, that no legitimate GraphViz/Diagrams/External-Data behavioral difference ever justified the old separator choice — confirmed by tracing all three `<graphviz>`-tag-providing extensions down to the same DOT source reaching the same `dot`/`neato` binary unmodified with respect to HTML-label content.

## Test plan

- [x] `composer test` (phpcs, phplint, MinusX, PHPUnit) passes with exit code 0 — 568 tests, 1372 assertions
- [x] New unit tests verified as genuine regression tests: reverting the fix (or removing the `<br />` argument at each call site) makes the corresponding test fail; confirmed for the field-value, `_wpg`-field-value, and header-wrap tests
- [x] New integration tests run against the real `dot` binary: confirmed a plain newline inside a DOT HTML label's `<td>` is swallowed by `dot` (words run together) while `<br />` correctly produces separate `<text>` lines in the rendered SVG
- [x] `testExternalDataGraphvizTagRendersCorrectLineBreaks` verified end-to-end: real `$wgExternalDataSources['graphviz']` config → `EDConnectorBase::loadConfig()` → real tag callback via `emulatedTags()` → wired onto a `Parser` via `setHook()` → parsed through `Parser::parse('<graphviz>...</graphviz>')` → real shell-out to `dot` → SVG output visually inspected and asserted
- [x] Confirmed the caption-fallback regression (introduced mid-refactor, caught via review) is real: reproduced the wrong output (`Team:Iota` instead of the caption) with the buggy intermediate code, then confirmed the fix and its regression test catch it
- [x] Confirmed both `_wpg`-field-value and header-wrap regression tests are environment-independent: since Diagrams happens to be loaded in the local dev container (making its fallback separator coincidentally also `<br />`), added explicit `ReflectionProperty` overrides forcing the non-Diagrams fallback, and verified (by temporarily removing the fix's `<br />` argument) that both tests fail correctly even with Diagrams loaded


alistair3149 (commented):
Nice fix — the diagnosis is right (a plain newline inside an HTML-label `<td>` really is swallowed by `dot`), and dropping the `str_replace( '<br />', '', $nodeTooltip )` quietly fixes a mangled tooltip as a bonus. A few things to address before merge.

**1. Fields-table header renders empty for a one-character caption** — `src/Graph/GraphFormatter.php:111-113`

`$label` now tests `!empty( $nodeLabel )` on the *raw* caption and then wraps unconditionally. `getWordWrappedText()` returns `''` for a single character (`\S.{1,24}(?=\s+|$)` needs ≥2 chars, `\S{25,}` needs ≥25), so the header cell comes out blank. Previously the caption was used verbatim on this path.

At defaults (`wordwraplimit=25`, `nodelabel=''`, `graphfields=yes`):

| caption | id | before | after |
|---|---|---|---|
| `H` | `H` | `[H]` | `[]` |
| `A` | `Page:A` | `[A]` | `[]` |

Reachable on any wiki with single-character titles (elements, gene symbols). Note that `w($caption) ?: w($id)` isn't the fix — for caption `A` / id `Page:A` it yields `Page:A` where before it rendered `A`. The root cause is `getWordWrappedText()` silently discarding tokens that match neither alternative (`getWordWrappedText( 'Value A', 5 ) === 'Value'`); fixing that, e.g. with `|\S+` as a final alternative, resolves both.

**2. The #816 fix isn't protected by any test in CI** — `tests/phpunit/Unit/Graph/GraphFormatterTest.php:472`

`testBuildGraphWordWrapsFieldValuesWithHtmlLineBreak()` is the only test covering the non-`_wpg` field-value call site (`GraphFormatter.php:152`) with a value long enough to wrap, and the comment at :477 argues it deliberately shouldn't pin `lineSeparator`. But CI installs Diagrams, so `lineSeparator === '<br />'` and the explicit argument is a no-op there.

Deleting `'<br />'` at :152 with Diagrams loaded: `GraphFormatterTest OK (27 tests)`, `GraphvizRenderingTest OK (6 tests)`. With Diagrams unloaded, the same mutation fails that test — so it only bites in the environment CI doesn't run.

Its two siblings (:531, :566) pin `PHP_EOL` by reflection. Adding the same three lines here closes it — or better, a data provider over `[ '<br />', PHP_EOL ]`, since "the explicit separator wins regardless of `lineSeparator`" is the actual invariant. `testFixedFormatterOutputRendersCorrectLineBreaks` in the integration file has the same hole.

**3. `@covers` on third-party classes breaks filtered runs** — `tests/phpunit/Integration/Graph/GraphvizRenderingTest.php:163, 240`

`@covers \EDConnectorBase::emulatedTags()` only resolves because `testExternalDataGraphvizTagRendersCorrectLineBreaks` registers an ad-hoc autoloader earlier in the run. With both extensions installed:

```
--filter GraphvizRenderingTest                   → OK (6 tests, 19 assertions)
--filter GraphvizRenderingTest::testValidCovers  → "@covers \EDConnectorBase::emulatedTags" is invalid
```

MediaWiki's inherited `testValidCovers` fails hard when a target can't be resolved, and the `markTestSkipped()` guards don't cover it since it's a separate test. Same applies to `@covers \MediaWiki\Extension\Diagrams\Dot::getSrc()` at :163. Simplest fix is to drop both annotations.

**4. Header wrapping changed beyond the separator**

Previously the header was wrapped only under `nodelabel=displaytitle`; with the default `nodelabel=''` the caption was emitted verbatim. Now both ternary arms wrap, so a 44-char caption goes from one line to `A Very Long Page Title<br />That Exceeds The Limit` and node geometry changes for existing `graphfields=yes` wikis on upgrade.

Defensible — an unwrapped header next to wrapped field rows is the real oddity — but worth saying so in RELEASE-NOTES (which currently reads as separator-only) and pinning with a test: `testBuildGraphWordWrapsFieldsTableHeaderWithHtmlLineBreak` inherits `BASE_PARAMS['nodelabel'] => 'displaytitle'`, and the other new header test uses a caption too short to wrap. Also worth aligning the no-fields branch, which still gates wrapping on `$useDisplayTitle` (:157-160) — the same caption now wraps in one branch but not the other.

**5. Pin ExternalData** — `extensions.local.json:10`

`"version": "master"` sits next to `"Diagrams": "1.0.0"` and is re-fetched on every `--no-cache` CI build, while the test depends on ED internals (`\EDConnectorBase::loadConfig()`, `emulatedTags()`, `EDConnectorExe::wikilinks4dot`) behind only an `is_file()` guard. Upstream has release tags through `3.6.1` plus `REL1_43`…`REL1_46` branches. Worth pinning, and promoting the guard to `class_exists( '\EDConnectorBase' )` so a moved class skips rather than errors.

**6. RELEASE-NOTES**

No `### Maintenance` entry for the new CI dependency (the Diagrams addition in d86581d1 set that precedent), and the tooltip fix — genuinely user-visible — isn't mentioned in the Bug Fixes line.

gesinn-it-gea:
Thanks again, @alistair3149 — this is another really thorough and helpful review. Addressed all six points in [3899b71](https://github.com/SemanticMediaWiki/SemanticResultFormats/commit/3899b7104fd87b2ed40115cd460b0fc3c31a85c3):

1. **Single-character caption bug** — `getWordWrappedText()`'s regex silently dropped tokens too short to match either alternative. Added a final `|\S+` alternative to catch them; added a regression test for a single-character caption.
2. **Test gap for the #816 fix itself** — consolidated the field-value/header wrap tests into a single `@dataProvider` over `['<br />', PHP_EOL]`, so they prove the explicit `'<br />'` argument wins regardless of `$this->lineSeparator`, not just coincidentally under whichever separator happens to be active in the running environment. Applied the same treatment to the header test, and switched it from the inherited `nodelabel=displaytitle` to `nodelabel=''`, since that's the case the #816 fix actually changed.
3. **`@covers` on third-party classes breaking `--filter` runs** — replaced both annotations with `@coversNothing` plus a comment explaining why, and verified the fix by running `--filter 'GraphvizRenderingTest::(testValidCovers|testDiagramsPreprocessingLeavesHtmlLabelBreaksUntouched)'` (excluding the test that registers ExternalData's autoloader) — passes now, reproduced the original failure first.
4. **Header wrapping changed beyond the separator** — you're right that it's defensible; documented it in RELEASE-NOTES and fixed the header test to actually exercise `nodelabel=''` (see #2). Left the no-fields branch's `$useDisplayTitle` gating alone: that path decides between a quoted-string and an HTML label depending on content (#846), which is a different correctness question than the header's — see the commit message for the fuller reasoning.
5. **Pin ExternalData** — pinned to `3.6.1` (latest release compatible with the CI matrix's MediaWiki 1.43–1.46), and added a `class_exists('\EDConnectorBase')` guard so a future rename skips cleanly instead of erroring.
6. **RELEASE-NOTES** — added the Maintenance entries, and one correction: your "dropping the `str_replace` quietly fixes a mangled tooltip" read — double-checked, and that removal is actually dead-code cleanup, not a functional fix. The caption feeding the tooltip is never wrapped on that path in the new code (it stays raw/unwrapped by construction), so it could never have contained a `<br />` to strip. Noted that accurately in RELEASE-NOTES instead.

Full `composer test` (phpcs, phplint, MinusX, PHPUnit) passes: 573 tests, 1380 assertions.

PR #1145, opened 2026-08-18
ArrayFormat: fix undeclared class error, make Arrays visible to Phan

### LINKED ISSUE #1144: Error in ArrayPrinter::createArray() when Arrays extension is installed — unqualified class reference resolves to wrong namespace
## Description

Using `format=array` with a `name=` parameter throws a fatal Error when the Arrays extension is actually installed:

```
Error: Class "SRF\ArrayFormat\ExtArrays" not found
```

## Root cause

`ArrayPrinter::createArray()` in `src/ArrayFormat/ArrayPrinter.php` calls the global `\ExtArrays` class unqualified:

```php
if ( defined( 'ExtArrays::VERSION' ) || class_exists( 'ExtArrays' ) ) {
    $parser = MediaWikiServices::getInstance()->getParser();
    ExtArrays::get( $parser )->createArray( $arrayId, $array );  // <-- missing leading \
    return true;
}
```

This file declares `namespace SRF\ArrayFormat;`. `class_exists('ExtArrays')` (string literal) correctly resolves to the global class, but the direct code reference `ExtArrays::get(...)` is resolved by PHP relative to the current namespace, so it looks for `SRF\ArrayFormat\ExtArrays` instead of the global `\ExtArrays` — which doesn't exist.

## Regression source

This is a regression from fb1a1e0e ("refactor(array,hash): move from formats/ to src/ and migrate to SRF\ namespace"). Before that commit the class had no namespace declaration, so the unqualified `ExtArrays::get(...)` call resolved correctly. The move into `SRF\ArrayFormat` silently broke this one reference. Phan didn't catch it because SRF's own `.phan/config.php` doesn't add the optional `Arrays` extension to its analysis scope (SRF deliberately doesn't require it) — so `ExtArrays` was entirely unknown to Phan and the finding was silently skipped rather than surfaced.

## Steps to reproduce

1. Install the Arrays extension alongside SRF.
2. Use `{{#ask: ... |format=array|name=someArrayName}}`.
3. Observe the fatal Error.

## Fix

A PR is attached that:
- Fixes the missing `\` (`\ExtArrays::get(...)`).
- Adds `Arrays` as a CI dependency via `extensions.local.json` and widens `.phan/config.php`'s scope to include it (matching the pattern already used for `SemanticMediaWiki` in SemanticDrilldown/KnowledgeGraph/PageForms), so this class of bug is caught by CI going forward instead of silently skipped.
- Adds a regression test that reproduces the exact error before the fix.

Note: `HashPrinter::createArray()` has the same unqualified-reference pattern for the separate `HashTables` extension's `ExtHashTables::get(...)` call — tracked separately in a follow-up issue.

gesinn-it-gea:
Fixes #1144

## Summary

- `ArrayPrinter::createArray()` calls the global `\ExtArrays` class unqualified while the file is namespaced (`SRF\ArrayFormat`). PHP resolves the unqualified static call relative to the current namespace, so it looks for `SRF\ArrayFormat\ExtArrays` instead of the global class — throwing an Error whenever the Arrays extension is actually installed. Regression from fb1a1e0e.
- Add `Arrays` as a CI dependency via `extensions.local.json` and widen `.phan/config.php` to include its source in Phan's scope (same pattern SemanticDrilldown/KnowledgeGraph/PageForms already use for `SemanticMediaWiki`), so Phan checks real usage against the dependency's actual API instead of silently skipping unknown classes.
- Regenerate `.phan/baseline.php` accordingly (only the `ArrayPrinter::createArray` entry for `PhanUndeclaredClassMethod` changes).

## Test plan

- Added `testCreateArrayCallsGlobalExtArraysWhenArraysExtensionInstalled`, which reproduces the exact production error before the fix (`Class "SRF\ArrayFormat\ExtArrays" not found`) and passes after it.
- Verified against SRF's own DCI container (MW 1.43, PHP 8.3, SMW 7.2.0 — the coverage/Phan matrix leg) with the real Arrays extension installed via `extensions.local.json`:
  - Phan: reproduces `PhanUndeclaredClassMethod ... \SRF\ArrayFormat\ExtArrays` before the fix, clean after.
  - Full PHPUnit suite: 574 tests, 1380 assertions, all green — no regressions.

### DIFF
--- a/.phan/baseline.php
+++ b/.phan/baseline.php
@@ -485,7 +485,7 @@ return [
             'PhanUndeclaredClass' => ['\\SRF\\ArrayFormat\\ArrayPrinter::__construct', '\\SRF\\ArrayFormat\\ArrayPrinter::getParamDefinitions', '\\SRF\\ArrayFormat\\ArrayPrinter::handleParameters'],
             'PhanUndeclaredClassConstant' => ['\\SRF\\ArrayFormat\\ArrayPrinter::createArray', '\\SRF\\ArrayFormat\\ArrayPrinter::getCfgSepText', '\\SRF\\ArrayFormat\\ArrayPrinter::getQueryMode'],
             'PhanUndeclaredClassInstanceof' => ['\\SRF\\ArrayFormat\\ArrayPrinter::getResultText'],
-            'PhanUndeclaredClassMethod' => ['\\SRF\\ArrayFormat\\ArrayPrinter::createArray', '\\SRF\\ArrayFormat\\ArrayPrinter::deliverPropertiesManyValues', '\\SRF\\ArrayFormat\\ArrayPrinter::getResultText'],
+            'PhanUndeclaredClassMethod' => ['\\SRF\\ArrayFormat\\ArrayPrinter::deliverPropertiesManyValues', '\\SRF\\ArrayFormat\\ArrayPrinter::getResultText'],
             'PhanUndeclaredConstant' => ['\\SRF\\ArrayFormat\\ArrayPrinter::deliverPropertiesManyValues'],
             'PhanUndeclaredExtendedClass' => ['src/ArrayFormat/ArrayPrinter.php'],
             'PhanUndeclaredMethod' => ['\\SRF\\ArrayFormat\\ArrayPrinter::getCfgSepText'],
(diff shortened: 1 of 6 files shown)


paladox:
Ci failed

gesinn-it-gea:
@paladox yeah, sorry, I was too fast with review request. Fix is landing soon

gesinn-it-gea:
@paladox all ci-legs are green now

PR #1147, opened 2026-08-19
fix(filtered): avoid null ParserOptions crash outside page-parse contexts

### LINKED ISSUE #802: Access to global `Parser` object can break in non-index.php-request contexts
Some result formats (e.g. "filtered") seem to rely on the global `Parser` object. Example: https://github.com/SemanticMediaWiki/SemanticResultFormats/blob/4.1.0/formats/filtered/src/Filtered.php#L102

This can lead to errors like
```
Error: Call to a member function getMaxIncludeSize() on null

Backtrace:

from includes/parser/Parser.php(2935)
#0 includes/parser/Parser.php(1609): Parser->replaceVariables()
#1 includes/parser/Parser.php(881): Parser->internalParse()
#2 extensions/SemanticResultFormats/formats/filtered/src/Filters/Filter.php(144): Parser->recursiveTagParse()
#3 extensions/SemanticResultFormats/formats/filtered/src/Filters/Filter.php(171): SRF\Filtered\Filter\Filter->addValueToJsConfig()
#4 extensions/SemanticResultFormats/formats/filtered/src/Filters/ValueFilter.php(38): SRF\Filtered\Filter\Filter->addValueListToJsConfig()
#5 extensions/SemanticResultFormats/formats/filtered/src/Filters/Filter.php(115): SRF\Filtered\Filter\ValueFilter->buildJsConfig()
#6 extensions/SemanticResultFormats/formats/filtered/src/Filtered.php(375): SRF\Filtered\Filter\Filter->getJsConfig()
#7 extensions/SemanticResultFormats/formats/filtered/src/Filtered.php(189): SRF\Filtered\Filtered->getFilterHtml()
#8 extensions/SemanticMediaWiki/src/Query/ResultPrinters/ResultPrinter.php(339): SRF\Filtered\Filtered->getResultText()
#9 extensions/SemanticMediaWiki/src/Query/ResultPrinters/ResultPrinter.php(304): SMW\Query\ResultPrinters\ResultPrinter->buildResult()
#10 extensions/SemanticMediaWiki/includes/query/SMW_QueryProcessor.php(348): SMW\Query\ResultPrinters\ResultPrinter->getResult()
#11 extensions/SemanticMediaWiki/src/ParserFunctions/AskParserFunction.php(370): SMWQueryProcessor::getResultFromQuery()
#12 extensions/SemanticMediaWiki/src/ParserFunctions/AskParserFunction.php(202): SMW\ParserFunctions\AskParserFunction->doFetchResultsFromFunctionParameters()
#13 extensions/SemanticMediaWiki/src/ParserFunctionFactory.php(402): SMW\ParserFunctions\AskParserFunction->parse()
#14 includes/parser/Parser.php(3437): SMW\ParserFunctionFactory->SMW\{closure}()
#15 includes/parser/Parser.php(3122): Parser->callParserFunction()
#16 includes/parser/PPFrame_Hash.php(275): Parser->braceSubstitution()
#17 includes/parser/Parser.php(2951): PPFrame_Hash->expand()
#18 includes/parser/Parser.php(1609): Parser->replaceVariables()
#19 includes/parser/Parser.php(723): Parser->internalParse()
...
```
when the code gets executed in a context different from the regular index.php-request.

Background:
In newer MediaWiki versions there is a dedicated `MediaWikiServices` service (`ParserFactory`) to create new, independent `Parser` objects

Example for such code:
```php
$page = $this->wikiPageFactory->newFromLinkTarget( $target );
$parserOptions = $page->makeParserOptions( 'canonical' );
$parser = $this->parserFactory->create();
$parserOutput = $parser->parse(
    $page->getContent()->getWikitextForTransclusion(),
    $page->getTitle(),
    $parserOptions,
    true,
    true,
    $page->getTitle()->getLatestRevID()
);
```
In this case the valid `ParserOptions` object passed to the `Parser::parse` method will not be set on the global `Parser` object. In the global object it may be `null`, depending on the context of this code (e.g. within a REST-API call).

One could probably consider this an edge case, but relying on global state objects should be avoided in general.

Usually this would be solved by passing the `Parser` object provided by MediaWiki to the handler implementation of the `{{#ask:}}` parser function (the callback registered by `Parser::setFunctionHook`) down to the result format. Unfortunately I didn't find a nice way to do that.

Any suggestions are welcome. If you can give me a hint about how to properly solve this, I can submit a patch.

gesinn-it-gea:
## Summary
- `Filtered::getParser()` returned the shared `Parser` service as-is, which is not initialized via `parse()`/`startExternalParse()` outside a regular page parse. Calling `recursiveTagParse()` on it (e.g. for filter parameters like `value filter switches`) then crashed with "Call to a member function getMaxIncludeSize() on null" in contexts such as `Special:Ask` or the REST API.
- `getParser()` now creates a dedicated `Parser` via `ParserFactory` and calls `startExternalParse()` on it when the shared instance has no `ParserOptions` set, leaving the shared service instance untouched for other consumers.

Fixes #802

## Test plan
- [x] Added `tests/phpunit/Integration/JSONScript/TestCases/filtered-04.json`: regression test reproducing `format=filtered` with a `value` filter switch parameter evaluated via `Special:Ask` (no prior page parse) — fails without the fix, passes with it.
- [x] Added `tests/phpunit/Unit/Filtered/ValueFilterTest.php`: unit-level reproduction using a freshly created `Parser` with unset `ParserOptions`.
- [x] Full `composer phpunit` suite run (576 tests) — no regressions; remaining failures/incomplete tests (`spreadsheet-01.json` intermittently, `HtmlValidator`-dependent tests) are pre-existing environment flakiness, reproduced independently of this change and passing in isolation.


gesinn-it-gea:
**Question: can the cached fallback parser conflict with the page parser once it gets initialized later?**

`getParser()` permanently caches its result in `$this->parser` (via `setParser()`). If the method is called early during a `{{#ask:}}` evaluation, while the shared parser service doesn't have `ParserOptions` yet, a dedicated ad-hoc parser is now created and initialized based on `RequestContext::getMain()->getTitle()` — and this ad-hoc parser stays cached for the lifetime of the `Filtered` instance.

If, later in the same request, the actual page parser (with the correct `Title`/`ParserOutput` of the page actually being parsed) does get initialized, subsequent `getParser()` calls would still return the already-cached ad-hoc parser instead. This matters in particular for `addConfigToOutput()` (Filtered.php:268), which uses `getParser()->getOutput()` to write the JS config into the `ParserOutput`.

Are there call sequences (e.g. nested `{{#ask:}}` calls via template transclusion) where `RequestContext::getMain()->getTitle()` doesn't match the title actually being parsed — and could that cause the JS config to end up attached to the wrong `ParserOutput`/page?

@alistair3149 @paladox @JeroenDeDauw what are your opinions on that?

